PR #19, opened 2024-06-28
fix(buy record): shift workflow of checking years and end timestamp t…

atticusofsparta:
…o not throw on permabuys

### DIFF
--- a/src/main.lua
+++ b/src/main.lua
@@ -78,7 +78,7 @@ local ActionMap = {
 }
 
 -- Write handlers
-Handlers.add(ActionMap.Transfer, utils.hasMatchingTag("Action", ActionMap.Transfer), function(msg)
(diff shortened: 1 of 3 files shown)


src/main.lua
```diff
@@ -78,7 +78,7 @@ local ActionMap = {
 }
 
 -- Write handlers
-Handlers.add(ActionMap.Transfer, utils.hasMatchingTag("Action", ActionMap.Transfer), function(msg)
+Handlers.add("info", utils.hasMatchingTag("Action", ActionMap.Transfer), function(msg)
```
dtfiedler, line 81:
```suggestion
Handlers.add(ActionMap.Transfer utils.hasMatchingTag("Action", ActionMap.Transfer), function(msg)
```

dtfiedler (approved):
please merge and apply to devnet and validate. once confirmed we can update testnet.

PR #31, opened 2024-08-02
PE-6533: Update process code to support ETH addresses

kunstmusik:


src/main.lua
```diff
@@ -83,9 +83,9 @@ local ActionMap = {
 Handlers.add(ActionMap.Transfer, utils.hasMatchingTag("Action", ActionMap.Transfer), function(msg)
 	-- assert recipient is a valid arweave address
 	local function checkAssertions()
-		assert(utils.isValidArweaveAddress(msg.Tags.Recipient), "Invalid recipient")
+		assert(utils.isValidAOAddress(msg.Recipient), "Invalid recipient")
 		assert(
-			tonumber(msg.Tags.Quantity) > 0 and utils.isInteger(tonumber(msg.Tags.Quantity)),
+			tonumber(msg.Quantity) > 0 and utils.isInteger(tonumber(msg.Quantity)),
```
dtfiedler, line 88: can you confirm Quantity is a higher level attribute on all messages?

kunstmusik: I'm actually a bit confused on when values should be read from msg.Tags and when they're read from msg directly. I noticed in the token blueprint it is reading Quantity and Recipient directly from msg:

https://cookbook_ao.g8way.io/guides/aos/blueprints/token.html

dtfiedler: I believe AO infra will append specific tags as higher-level attributes to all messages. I believe they do so for `Quantity`, `From`, `Recepient` but it doesn't show anywhere in the spec - https://cookbook_ao.ar-io.dev/concepts/messages.html

hence why I've always deferred to using Tags as that's what they are on the actual data item

Ref: https://ao.ar.io/#/read

image

kunstmusik: I still find it ambiguous, but to be safe I've normalized back to msg.Tags.Quantity and msg.Tags.Recipient

src/utils.lua
```diff
@@ -131,6 +131,45 @@ function utils.isValidArweaveAddress(address)
...
+		local hexCharValue = tonumber(hexChar, 16)
+		local char = string.sub(hex, i, i)
+		if hexCharValue > 7 then
+			char = string.upper(char)
+		end
+		checksumAddress = checksumAddress .. char
+	end
+
+	return checksumAddress
+end
```
dtfiedler, line 164: nice!

test/arns.test.js
```diff
@@ -30,38 +35,62 @@ describe('ArNS', async () => {
...
+        processId: ''.padEnd(43, 'a'),
+        purchasePrice: 600000000,
+        startTimestamp: buyRecordData.startTimestamp,
+        endTimestamp: buyRecordData.endTimestamp,
+        type: 'lease',
+        undernameLimit: 10,
+      });
+    };
+    await runBuyRecord(STUB_ADDRESS);
+    await runBuyRecord(testEthAddress);
```
dtfiedler, line 93: nit - maybe worth putting separate `it` blocks and pulling up `runBuyRecord`

test/arns.test.js
```diff
@@ -155,36 +184,64 @@ describe('ArNS', async () => {
...
-        Tags: [
-          { name: 'Action', value: 'Record' },
-          { name: 'Name', value: 'test-name' },
-        ],
-      },
-      increaseUndernameResult.Memory,
-    );
-    const record = JSON.parse(result.Messages[0].Data);
-    assert.equal(record.undernameLimit, 11);
+    const runIncreaseUndername = async (sender) => {
```
dtfiedler, line 187: nit: consider changing to `assertIncreaseUndername` given these function contains the assertions

dtfiedler:
Looks good! Thank you for the thorough tests. Merging to develop will trigger the CI/CD pipeline to evolve the devnet process.

dtfiedler (approved):

PR #32, opened 2024-08-02
chore(ci/cd): add evolve ci

atticusofsparta:
Addresses PE-6325


.github/workflows/evolve-prod.yml
```diff
@@ -0,0 +1,23 @@
+name: Evolve Devnet ANT Registry
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  # TODO: add unit tests using bused
+  evolve:
```
dtfiedler, line 10: Let's add this to our build step and require tests to pass (i.e. put it all in one file)

dtfiedler (changes requested):

dtfiedler (approved):

PR #49, opened 2024-08-30
fix(tick): prune state on every message

dtfiedler:
We can additional add the "distribute" logic to this so rewards are to be distributed at the nearest message. We will test this before we make that update.


dtfiedler:
going to add a test that validates the prepend handler works as expected

PR #57, opened 2024-09-10
feat(gar): slash operators by 20% of minimum gateway stake if they fail 30 consecutive epochs

dtfiedler:


src/gar.lua
```diff
@@ -89,17 +90,17 @@ function gar.leaveNetwork(from, currentTimestamp, msgId)
 
 	-- Add minimum staked tokens to a vault that unlocks after the gateway completely leaves the network
 	gateway.vaults[from] = {
-		balance = gar.getSettings().operators.minStake,
+		balance = math.min(gar.getSettings().operators.minStake, gateway.operatorStake),
```
arielmelendez, line 93: Is the system in an invariant state if `operatorStake` is less than `minStake` here? If so, hopefully we're handling that elsewhere and the min should not be necessary; you'd be able to just use `minStake`.

src/gar.lua
```diff
@@ -628,7 +629,9 @@ function gar.pruneGateways(currentTimestamp, msgId)
 				and garSettings ~= nil
 				and gateway.stats.failedConsecutiveEpochs >= garSettings.operators.failedEpochCountMax
 			then
-				-- mark as leaving
+				-- slash operator stake and return 20% of operator stake to the protocol balance and mark as leaving
+				local slashAmount = math.floor(gateway.operatorStake * garSettings.operators.failedEpochSlashPercentage)
```
arielmelendez, line 633: Is choosing between floor and ceil here all that consequential?

src/gar.lua
```diff
@@ -640,6 +643,25 @@ function gar.pruneGateways(currentTimestamp, msgId)
...
+	end
+	local garSettings = gar.getSettings()
+	if garSettings == nil then
+		error("Gateway Registry settings do not exist")
+	end
+
+	gateway.operatorStake = gateway.operatorStake - slashAmount
+	balances.increaseBalance(ao.id, slashAmount)
+	GatewayRegistry[address] = gateway
+	-- TODO: send slash notice to gateway address
```
arielmelendez, line 662: Are we leaving this todo for another ticket?

arielmelendez (approved):
Approved with some questions.

PR #62, opened 2024-09-16
feat(PE-6683): add returning fill price list for all name lengthsif no name provided

atticusofsparta:


tests/arns.test.mjs
```diff
@@ -263,6 +263,24 @@ describe('ArNS', async () => {
     assert.equal(tokenCost, 600000000);
   });
 
+  it('should get registration prices', async () => {
+    const priceListResult = await handle({
+      Tags: [
+        { name: 'Action', value: 'Get-Registration-Prices' },
+        { name: 'Intent', value: 'Buy-Record' },
```
dtfiedler, line 270: wondering if we need intent - and should it be `GET-REGISTRATION-FEES`

atticusofsparta: Removed the intent and changes to Get-Registration-Fees

spec/arns_spec.lua
```diff
@@ -3,6 +3,7 @@ local constants = require("constants")
 local arns = require("arns")
 local balances = require("balances")
 local demand = require("demand")
+local json = require("json")
```
dtfiedler, line 6:
```suggestion
```

PR #65, opened 2024-09-20
PE-6794: IO Usage Events

arielmelendez:
Added tracking events for:
- Transfer
- Create-Vault
- Delegate-Stake
- Decrease-Delegate-Stake
- Cancel-Delegate-Withdrawl
- "Extend-Vault"
- "Increase-Vault"

Added function `eventingPcall` that DRYs out sending errors along with events when pcalls fail.

### DIFF
--- a/src/ao_event.lua
+++ b/src/ao_event.lua
@@ -1,107 +1,114 @@
 -- Factory function for creating an "AOEvent"
 local function AOEvent(initialData)
-  local event = {
-    sampleRate = nil, -- Optional sample rate
-  }
+	local event = {
+		sampleRate = nil, -- Optional sample rate
+	}
 
-  if type(initialData) ~= "table" then
-    print("ERROR: AOEvent data must be a table.")
-    event.data = {}
-  else
-    event.data = initialData
-  end
+	if type(initialData) ~= "table" then
+		print("ERROR: AOEvent data must be a table.")
+		event.data = {}
+	else
+		event.data = initialData
+	end
 
-  local function isValidType(value)
-    local valueType = type(value)
-    return valueType == "string" or valueType == "number" or valueType == "boolean" or value == nil
-  end
+	local function isValidType(value)
+		local valueType = type(value)
+		return valueType == "string" or valueType == "number" or valueType == "boolean" or value == nil
+	end
 
-  function event:addField(key, value)
-    if type(key) ~= "string" then
-      print("ERROR: Field key must be a string.")
-      return self
-    end
-    if not isValidType(value) then
-      print("ERROR: Invalid field value type: " .. type(value) .. ". Supported types are string, number, boolean, or nil.")
-      return self
-    end
-    self.data[key] = value
-    return self
-  end
+	function event:addField(key, value)
+		if type(key) ~= "string" then
+			print("ERROR: Field key must be a string.")
+			return self
+		end
+		if not isValidType(value) then
+			print(
+				"ERROR: Invalid field value type: "
+					.. type(value)
+					.. ". Supported types are string, number, boolean, or nil."
+			)
+			return self
+		end
+		self.data[key] = value
+		return self
+	end
 
-  function event:addFields(fields)
-    if type(fields) ~= "table" then
-      print("ERROR: Fields must be provided as a table.")
-      return self
-    end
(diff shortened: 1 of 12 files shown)


atticusofsparta (commented):
Reccomend adding integration tests for events using AoLoader

src/main.lua
```diff
@@ -193,40 +230,53 @@ Handlers.add(ActionMap.CreateVault, utils.hasMatchingTag("Action", ActionMap.Cre
...
+		ioEvent:addField("VaultEndTimestamp", vault.endTimestamp)
 	end
+
+	ao.send({
+		Target = msg.From,
+		Tags = {
+			Action = "Vault-Created-Notice",
+			["Vault-Id"] = msg.Id,
+		},
+		Data = json.encode(shouldContinue2),
```
arielmelendez, line 277: This should prooooobably be `vault` (json encoded).

src/main.lua
```diff
@@ -292,34 +344,49 @@ Handlers.add(ActionMap.ExtendVault, utils.hasMatchingTag("Action", ActionMap.Ext
...
 			Target = msg.From,
 			Tags = { Action = "Invalid-Extend-Vault-Notice", Error = "Invalid-Extend-Vault" },
-			Data = tostring(err),
-		})
-	else
-		ao.send({
-			Target = msg.From,
-			Tags = { Action = "Vault-Extended-Notice" },
-			Data = json.encode(result),
+			Data = tostring(error),
```
arielmelendez, line 364: `error` should probably be json encoded `vault`.

src/main.lua
```diff
@@ -693,87 +775,119 @@ Handlers.add(ActionMap.DelegateStake, utils.hasMatchingTag("Action", ActionMap.D
...
-			ao.send({
-				Target = msg.From,
-				Tags = {
-					Action = "Invalid-Cancel-Delegate-Withdrawl-Notice",
-					Error = "Invalid-Cancel-Delegate-Withdrawl",
-				},
-				Data = tostring(result),
-			})
-		else
+		local shouldContinue2, result = eventingPcall(ioEvent, function(error)
```
arielmelendez, line 846: Rename `result` to `gateway`.

src/main.lua
arielmelendez: 9e21524

src/ao_event.lua
```diff
@@ -0,0 +1,115 @@
+-- Factory function for creating an "AOEvent"
+local function AOEvent(initialData)
```
dtfiedler, line 2: i haven't seen lua classes created this way - the most common pattern i've seen is more like

```
-- Define a class
local MyClass = {}
MyClass.__index = MyClass

-- Constructor for the class
function MyClass:new(name, age)
    local instance = setmetatable({}, MyClass) -- Set the metatable for the new object
    instance.name = name
    instance.age = age
    return instance
end

-- A method for the class
function MyClass:greet()
    print("Hello, my name is " .. self.name .. " and I am " .. self.age .. " years old.")
end

-- Create an instance of the class
local person = MyClass:new("John", 30)
person:greet()  -- Output: Hello, my name is John and I am 30 years old.
```

but regardless - we can look to use classes more for certain objects (gateways, records) to put some protections around accessing/modifying properties directly)

src/gar.lua
```diff
@@ -705,6 +706,10 @@ function gar.cancelDelegateWithdrawal(from, gatewayAddress, vaultId)
 	delegate.delegatedStake = delegate.delegatedStake + vaultBalance
 	gateway.totalDelegatedStake = gateway.totalDelegatedStake + vaultBalance
 	GatewayRegistry[gatewayAddress] = gateway
+	return {
+		delegate = gar.getGateway(gatewayAddress).delegates[from],
+		totalDelegatedStake = gateway.totalDelegatedStake
+	}
```
dtfiedler, line 712: inconsequential change - but required a modification to the unit test that validates the result is no linger `nil`

dtfiedler (approved):
can get this into devnet and test against a local cu/honeycomb - if there's an issue, our CI/CD and monitoring tests https://github.com/ar-io/ar-io-network-process/actions/workflows/monitor.yaml

should catch broken handlers/unexpected state

PR #66, opened 2024-09-23
PE-6667 - add bundling services into the gar

vilenarios:
Enables gateways to advertise any dedicated bundlers they may be using. We do not do any verification of these bundlers, so users should verify themselves before spending/paying for any service.


src/main.lua
```diff
@@ -828,6 +845,22 @@ Handlers.add(
 			return
 		end
 
+		local services = nil
+        if msg.Tags.Services then
```
arielmelendez, line 849: Nit: some funny tabbing here. Plz run stylua if you can.

vilenarios: Ran stylua and fixed the funky tabs and missing commas.

PR #68, opened 2024-09-24
PE-6795: Instrument record-related events

arielmelendez:
Instrumentation for:
- Buy-Record
- Extend-Lease
- Increase-Undername-Limit
- Record pruned (name expired)

### DIFF
--- a/src/arns.lua
+++ b/src/arns.lua
@@ -430,7 +430,9 @@ function arns.assertValidIncreaseUndername(record, qty, currentTimestamp)
 end
 
 function arns.removeRecord(name)
+	local record = NameRegistry.records[name]
 	NameRegistry.records[name] = nil
+	return record
 end
 
 function arns.removeReservedName(name)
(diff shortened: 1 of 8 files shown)


src/arns.lua
```diff
@@ -99,7 +107,16 @@ function arns.extendLease(from, name, years, currentTimestamp)
...
-	return arns.getRecord(name)
+	return {
+		record = arns.getRecord(name),
+		totalExtensionFee = totalExtensionFee,
+		baseRegistrationFee = baseRegistrationFee,
+		remainingBalance = balances.getBalance(from),
+		protocolBalance = balances.getBalance(ao.id),
+		recordsCount = utils.lengthOfTable(NameRegistry.records),
+		reservedRecordsCount = utils.lengthOfTable(NameRegistry.reserved),
+		df = demand.getDemandFactorInfo(),
```
dtfiedler, line 118: do these need to be returned by this function? you can get them directly in main.lua

arielmelendez: Sure I can make that change :)

src/main.lua
```diff
@@ -471,26 +517,28 @@ Handlers.add(ActionMap.BuyRecord, utils.hasMatchingTag("Action", ActionMap.BuyRe
...
+		record = result.record
+		addRecordResultFields(ioEvent, result)
+	end
+
+	ao.send({
+		Target = msg.From,
+		Tags = { Action = "Buy-Record-Notice", Name = msg.Tags.Name },
+		Data = json.encode(record),
+	})
+	ioEvent:printEvent()
```
dtfiedler, line 535: i don't think it's supported, but it would be nice if we could add a `postpend` handler that always prints this IO event before returning.

arielmelendez: Was thinking the same...

dtfiedler (approved):

PR #71, opened 2024-09-26
fix(handlers): replace `tick` with `prune` prepended handler

dtfiedler:
`tick` existed before we moved it to a prepended handler, which means it did not get updated in the `Handlers.list` and only executed before any other handlers it was placed before (e.g. `paginedGateways, Epochs, etc.) - here we replace tick with `prune` and then we will have to manually remove the legacy `tick` handler to avoid duplicate processing.


arielmelendez (approved):

tests/monitor/monitor.test.mjs
```diff
@@ -36,6 +36,30 @@ describe('setup', () => {
...
+        '_eval should be first handler',
+      );
+      assert.ok(
+        handlersList.indexOf('_default') === 1,
+        '_default should be second handler',
+      );
+      assert.ok(
+        handlersList.indexOf('prune') === 2,
+        'prune should be third handler',
+      );
```
dtfiedler, line 55: these will fail until we get clarification from AO team if the following handlers are still necessary after upgrading to aos 2.0.0

```
{
  {
     handle = function: 0x912128,
     pattern = function: 0xd37708,
     name = "Assignment-Check"
  },
  {
     handle = function: 0x755910,
     pattern = function: 0x929600,
     name = "sec-patch-6-5-2024"
  },
```

dtfiedler: image

dtfiedler: https://discord.com/channels/1210396395643863101/1210403116361580574/1288924014718357514

PR #77, opened 2024-09-27
fix(prune): copy epochs and restore them if prune fails, remove max delegates for gateways

dtfiedler:


src/main.lua
```diff
@@ -130,13 +130,14 @@ end, function(msg)
...
-	-- TODO: we should copy state here and restore if tick fails, but that requires larger memory - DO NOT DO THIS UNTIL WE START PRUNING STATE of epochs and distributions
-	local status, resultOrError = pcall(tick.pruneState, msgTimestamp, msgId)
 	local previousState = {
 		Vaults = utils.deepCopy(Vaults),
 		GatewayRegistry = utils.deepCopy(GatewayRegistry),
 		NameRegistry = utils.deepCopy(NameRegistry),
+		Epochs = utils.deepCopy(Epochs),
+		-- TODO: add vaults and balances
 	}
+	local status, resultOrError = pcall(tick.pruneState, msgTimestamp, msgId)
```
arielmelendez, line 140: I'm glad you fixed this because I was confused why we'd stash the previous state AFTER we already made the change 😅

arielmelendez (approved):

PR #85, opened 2024-10-01
feat(events): instrument Increase/Decrease-Operator-Stake PE-6854

arielmelendez:


src/main.lua
```diff
@@ -838,31 +838,43 @@ Handlers.add(
...
 			})
+		end, gar.increaseOperatorStake, msg.From, quantity)
+		if not shouldContinue2 then
+			return
 		end
+
+		msg.ioEvent:addField("Sender-New-Balance", balances[msg.From])
+		if gateway ~= nil then
+			msg.ioEvent:addField("GW-New-Operator-Stake", gateway.operatorStake)
+			msg.ioEvent:addField("GW-Previous-Operator-Stake", gateway.operatorStake - quantity)
```
dtfiedler, line 869: nit: add
```
msg.ioEvent:addField(`Gateway`, msg.From) -- or add `gateway.gatewayAddress` as an attribute on the returned gateway object
```
and remove the `GW` prefex in the provided tags

src/main.lua
```diff
@@ -877,31 +889,54 @@ Handlers.add(
...
+		if not shouldContinue2 then
+			return
 		end
+
+		msg.ioEvent:addField("Sender-New-Balance", balances[msg.From]) -- should be unchanged
+		if gateway ~= nil then
+			local previousStake = gateway.operatorStake + quantity
+			msg.ioEvent:addField("GW-New-Operator-Stake", gateway.operatorStake)
+			msg.ioEvent:addField("GW-Previous-Operator-Stake", previousStake)
+			msg.ioEvent:addField("GW-Vaults-Count", #(gateway.vaults or {}))
```
dtfiedler, line 922: see same comment as above

src/main.lua
arielmelendez: From is auto-added to every event. Would it be important to you to see it also added to a `Gateway` field?

src/main.lua
arielmelendez: `GW-Vaults-Count` is a holdover from Leave-Network where multiple kinds of vault data is returned. In fairness, they're all gateway vaults (as opposed to delegate vaults). This also distinguishes against ordinary vaults that might be associated with the wallet holder. I can remove the GW- prepend from the operator stake ones though.

src/main.lua
arielmelendez: a5edd23

src/main.lua
arielmelendez: a5edd23

PR #88, opened 2024-10-03
PE-6791: Send arrays as as arrays rather than delimited lists

arielmelendez:
- Collect event info for every handler by using a wrapper function
- No longer print events in eventingPcall; expect wrapper function to handle
- Repair collection of event data for a few tables

### DIFF
--- a/src/main.lua
+++ b/src/main.lua
@@ -88,7 +88,6 @@ local function eventingPcall(ioEvent, onError, fnToCall, ...)
 	if not status then
 		onError(result)
 		ioEvent:addField("Error", result)
-		ioEvent:printEvent()
 		return status
 	end
 	return status, result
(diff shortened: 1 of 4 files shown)


dtfiedler (approved):

src/main.lua
```diff
@@ -130,6 +128,15 @@ local function gatewayStats()
...
 end
 
+local function addEventingHandler(handlerName, pattern, handleFn)
+	Handlers.add(handlerName, pattern, function(msg)
+		eventingPcall(msg.ioEvent, function()
+			-- No op
+		end, handleFn, msg)
+		msg.ioEvent:printEvent()
+	end)
+end
```
dtfiedler, line 138: 👏

tests/arns.test.mjs
```diff
@@ -63,6 +63,22 @@ describe('ArNS', async () => {
...
+    
+    // Fields based on timestamps will be unstable for testing, but verify their type
+    assert.equal("number", typeof buyRecordEvent.Timestamp);
+    assert.equal("number", typeof buyRecordEvent['Epoch-Index']);
+    assert.equal("number", typeof buyRecordEvent['Start-Timestamp']);
+    assert.equal("number", typeof buyRecordEvent['End-Timestamp']);
+    delete buyRecordEvent.Timestamp;
+    delete buyRecordEvent['Epoch-Index'];
+    delete buyRecordEvent['Start-Timestamp'];
+    delete buyRecordEvent['End-Timestamp'];
```
dtfiedler, line 76: these are all controllable in the test setup, likely don't need to delete and node does support things like `assert.isNumber()` if we don't care about specific values

arielmelendez: The values are just unstable in testing environment since they're based on the current timestamp. Thanks for the tip on isNumber

PR #92, opened 2024-10-09
feat(PE-6656) - return initiated dutch auctions

dtfiedler:
Summary:
Introduces and `Auction` object class that is used to store auctions on the `NameRegistry`. We should look towards using more of these classes to leverage some better static analysis and type enforcement on functions. The class is responsible for computing prices based on the initialization details (baseFee, demandFactor, etc.). Auctions are stored as instances of this class, and returned as simple table (excluding function) so clients can compute prices themselves.

This initial implementation includes handler support for ANTs returning `permabuy` name registrations. In a separate PR we'll introduce Lease-Expiration-Intiated-Auctions where a name will go in to auction after the grace period has ended.

Checklist:
- [x] business logic (and docs) to create and store auctions in state with pricing information
- [x] handler for `Release-Name` that allows integration with ANTs
- [x] handler for `Auction-Info` to get auction information of a given name
- [x] handler for `Auction-Bid` to submit bid to acquire name in auction and update arns registry
- [x] unit tests with >80% coverage
- [x] integration tests for all handlers
- [x] update ANT source code to support `Release-Name` handler that forwards to this process
    - PR: https://github.com/ar-io/ar-io-ant-process/pull/23
- [x] update [ar-io-sdk](https://github.com/ar-io/ar-io-sdk) to include auction types and APIs for auctions
    - PR: https://github.com/ar-io/ar-io-sdk/pull/235
- [ ] add event data for auction creation and auction bids
    - WILL DO IN A SEPARATE PR ONCE E2E TESTING IN DEVNET HAS BEEN VALIDATED

TODOs:
- [x] storing prices in auction state impacts the handler response (`Data` ends up being an empty table - not sure if this is a bug in AO or we are hitting a memory limit). it also costs ~600KiB to store those values in state, which could bloat our memory when several auctions are in process. we could consider just storing the settings of the auction pricing in state, and use that to calculate prices when necessary (still have to create the table, but response doesn't need to contain it). need to think more about this and what is necessary for clients to be able to showing pricing charts.
    - we've modified to store the auction settings on the auction object, and then clients will compute the price with various parameters

Here is the sequence diagram for a permabuy initiated auction
image

### DIFF
--- a/spec/arns_spec.lua
+++ b/spec/arns_spec.lua
@@ -644,7 +644,7 @@ describe("arns", function()
 					local auctionIntervalMs = 1000 * 60 * 2 -- ~2 min per price interval
 					local bidTimestamp = startTimestamp + auctionIntervalMs - 1 -- 1 ms before the next price interval
 					local auction = arns.createAuction("test-name", "permabuy", startTimestamp, "test-initiator")
-					local wonAuctionRecord = arns.submitAuctionBid(
+					local result = arns.submitAuctionBid(
 						"test-name",
 						auction.startPrice,
 						testAddressArweave,
@@ -665,7 +665,7 @@ describe("arns", function()
 					assert.are.equal(balances[_G.ao.id], expectedPrice * 0.5)
 					assert.are.equal(NameRegistry.auctions["test-name"], nil)
 					assert.same(expectedRecord, NameRegistry.records["test-name"])
-					assert.same(expectedRecord, wonAuctionRecord)
+					assert.same(expectedRecord, result.record)
 				end
 			)
 
--- a/spec/arns_spec.lua
+++ b/spec/arns_spec.lua
@@ -602,26 +602,6 @@ describe("arns", function()
 				assert.are.equal(NameRegistry.records["test-name"], nil)
 			end)
 
-			it("should return the correct price for an auction at a given timestamp", function()
-				local startTimestamp = 1000000
-				local auction = arns.createAuction("test-name", startTimestamp, "test-initiator")
-				local expectedEndTimestamp = startTimestamp + 1000 * 60 * 60 * 24 * 14 -- 14 days
-				local auctionIntervalMs = 1000 * 60 * 2 -- ~2 min per price interval
-				local expectedPrices = {}
-				-- check timestamp at every interval
-				for timestampAtInterval = startTimestamp, expectedEndTimestamp - auctionIntervalMs, auctionIntervalMs do
-					local intervalsSinceStart = math.floor((timestampAtInterval - startTimestamp) / auctionIntervalMs)
-					local totalDecaySinceStart = math.min(1, 0.000002 * intervalsSinceStart)
-					local priceAtInterval = math.floor(auction.startPrice * ((1 - totalDecaySinceStart) ^ 190))
-					expectedPrices[timestampAtInterval] = priceAtInterval
-					-- assert the price is correct at that timestamp
-					assert.are.equal(priceAtInterval, auction.prices[timestampAtInterval])
-				end
-				-- ensure the full object matches and # of prices is correct
-				assert.are.equal(utils.lengthOfTable(expectedPrices), utils.lengthOfTable(auction.prices))
-				assert.are.same(expectedPrices, auction.prices)
-			end)
-
 			it("should throw an error if the name is already in the auction map", function()
 				_G.NameRegistry.auctions = {
 					["test-name"] = {
@@ -650,7 +630,7 @@ describe("arns", function()
 			it("should return the auction", function()
 				local auction = arns.createAuction("test-name", 1000000, "test-initiator")
 				local retrievedAuction = arns.getAuction("test-name")
-				assert.are.equal(auction, retrievedAuction)
+				assert.are.same(retrievedAuction, auction)
 			end)
 		end)
 
@@ -658,9 +638,15 @@ describe("arns", function()
 			it("should return the correct price for an auction at a given timestamp", function()
 				local startTimestamp = 1000000
 				local auction = arns.createAuction("test-name", startTimestamp, "test-initiator")
-				local randomTimestampDuringAuction = startTimestamp + 1000 * 60 * 60 * 24 * 7 -- 1 week into the auction
-				local priceAtRandomTimestamp = arns.getCurrentBidPriceForAuction(auction, randomTimestampDuringAuction)
-				assert.are.equal(priceAtRandomTimestamp, auction.prices[randomTimestampDuringAuction])
+				local currentTimestamp = startTimestamp + 1000 * 60 * 60 * 24 * 7 -- 1 week into the auction
+				local decayRate = 0.00000002
+				local scalingExponent = 190
+				local timeSinceStart = currentTimestamp - auction.startTimestamp
+				local totalDecaySinceStart = decayRate * timeSinceStart
+				local expectedPriceAtTimestamp =
+					math.floor(auction.startPrice * ((1 - totalDecaySinceStart) ^ scalingExponent))
+				local priceAtTimestamp = arns.getCurrentBidPriceForAuction(auction, currentTimestamp)
+				assert.are.equal(priceAtTimestamp, expectedPriceAtTimestamp)
 			end)
 		end)
 
@@ -669,8 +655,7 @@ describe("arns", function()
 				"should accept bid on an existing auction and transfer tokens to the auction initiator and protocol balance, and create the record",
 				function()
 					local startTimestamp = 1000000
-					local auctionIntervalMs = 1000 * 60 * 2 -- ~2 min per price interval
-					local bidTimestamp = startTimestamp + auctionIntervalMs - 1 -- 1 ms before the next price interval
+					local bidTimestamp = startTimestamp + 1000 * 60 * 2 -- 2 min into the auction
 					local demandBefore = demand.getCurrentPeriodPurchases()
 					local revenueBefore = demand.getCurrentPeriodRevenue()
 					local auction = arns.createAuction("test-name", startTimestamp, "test-initiator")
@@ -682,7 +667,8 @@ describe("arns", function()
 						"test-process-id"
 					)
 					local balances = balances.getBalances()
-					local expectedPrice = auction.startPrice -- bid submitted before next price interval
+					local expectedPrice =
+						math.floor(auction.startPrice * ((1 - (0.00000002 * (bidTimestamp - startTimestamp))) ^ 190))
 					local expectedRecord = {
 						endTimestamp = nil,
 						processId = "test-process-id",
@@ -691,13 +677,23 @@ describe("arns", function()
 						type = "permabuy",
 						undernameLimit = 10,
 					}
-					assert.are.equal(balances["test-initiator"], expectedPrice * 0.5)
-					assert.are.equal(balances[_G.ao.id], expectedPrice * 0.5)
-					assert.are.equal(NameRegistry.auctions["test-name"], nil)
-					assert.same(expectedRecord, NameRegistry.records["test-name"])
-					assert.same(expectedRecord, result.record)
-					assert.are.equal(demand.getCurrentPeriodPurchases(), demandBefore + 1)
-					assert.are.equal(demand.getCurrentPeriodRevenue(), revenueBefore + expectedPrice)
+					local expectedInitiatorReward = math.floor(expectedPrice * 0.5)
+					local expectedProtocolReward = expectedPrice - expectedInitiatorReward
+					assert.are.equal(expectedInitiatorReward, balances["test-initiator"])
+					assert.are.equal(expectedProtocolReward, balances[_G.ao.id])
+					assert.are.equal(nil, NameRegistry.auctions["test-name"])
+					assert.are.same(expectedRecord, NameRegistry.records["test-name"])
+					assert.are.same(expectedRecord, result.record)
+					assert.are.equal(
+						demandBefore + 1,
+						demand.getCurrentPeriodPurchases(),
+						"Purchases should increase by 1"
+					)
+					assert.are.equal(
+						revenueBefore + expectedPrice,
+						demand.getCurrentPeriodRevenue(),
+						"Revenue should increase by the bid amount"
+					)
 				end
 			)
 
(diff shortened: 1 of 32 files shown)


arielmelendez:
General request to add events for this flow when the PR is approaching readiness for review.

spec/arns_spec.lua
```diff
@@ -606,6 +643,227 @@ describe("arns", function()
...
+						"Price at timestamp" .. i .. " should be " .. expectedPriceAtTimestamp
+					)
+				end
+				-- make sure the last price at the end of the auction is the floor price
+				local lastProvidedPrice = prices[auction.endTimestamp]
+				local lastComputedPrice = auction:getPriceForAuctionAtTimestamp(auction.endTimestamp, "lease", 1)
+				local listPricePercentDifference = (lastComputedPrice - lastProvidedPrice) / lastProvidedPrice
+				assert.is_true(
+					listPricePercentDifference <= 0.0001,
+					"Last price should be within 0.01% of the final price in the interval. Last computed: "
```
atticusofsparta, line 755:
```suggestion
					"Last price should be within 0.0001% of the final price in the interval. Last computed: "
```

atticusofsparta, line 755:
```suggestion
					"Last price should be within 0.001% of the final price in the interval. Last computed: "
```

src/arns.lua
```diff
@@ -278,7 +279,7 @@ function arns.assertValidBuyRecord(name, years, purchaseType, processId)
 	assert(name:match("^%w") and name:match("%w$") and name:match("^[%w-]+$"), "Name pattern is invalid.")
```
atticusofsparta, line 279: should we have a util for this?

src/arns.lua
```diff
@@ -423,14 +424,106 @@ function arns.assertValidIncreaseUndername(record, qty, currentTimestamp)
 	return true
 end
 
+-- AUCTIONS
+
+--- Creates an auction for a given name
+--- @param name string The name of the auction
+--- @param timestamp number The timestamp to start the auction
+--- @param initiator string The address of the initiator of the auction
+--- @return Auction|nil The auction instance
```
atticusofsparta, line 433: 👏

src/arns.lua
```diff
@@ -423,14 +424,106 @@ function arns.assertValidIncreaseUndername(record, qty, currentTimestamp)
...
+	return auction
+end
+
+function arns.getAuction(name)
+	return NameRegistry.auctions[name]
+end
+
+function arns.getAuctions()
+	return NameRegistry.auctions or {}
+end
```
atticusofsparta, line 457: the `or {}` implies auctions could be undefined which would indicate that `NameRegistry.auctions[name]` could attempt to index a nil value

dtfiedler: it can be, if we don't update the state manually

src/arns.lua
dtfiedler: eventually

spec/arns_spec.lua
dtfiedler: negative - 0.0001 = 0.01%

atticusofsparta: Ah, my bad misread

spec/arns_spec.lua
```diff
@@ -606,6 +643,227 @@ describe("arns", function()
...
+
+			it("should throw an error if the name is not registered", function()
+				_G.NameRegistry.records["test-name"] = nil
+				local status, error = pcall(arns.createAuction, "test-name", 1000000, "test-initiator")
+				assert.is_false(status)
+				assert.match("Name is not registered", error)
+			end)
+		end)
+
+		describe("getAuction", function()
```
arielmelendez, line 694: Could add a test for a missing auction

src/arns.lua
```diff
@@ -30,14 +32,14 @@ function arns.buyRecord(name, purchaseType, years, from, timestamp, processId)
 		error("Insufficient balance")
 	end
 
-	if
-		arns.getRecord(name) ~= nil and arns.getRecord(name).endTimestamp == nil
-		or arns.getRecord(name) ~= nil and arns.getRecord(name).endTimestamp + constants.gracePeriodMs > timestamp
-	then
+	local record = arns.getRecord(name)
+	local isPermabuy = record ~= nil and record.type == "permabuy"
+	local isActiveLease = record ~= nil and (record.endTimestamp or 0) + constants.gracePeriodMs > timestamp
```
arielmelendez, line 37: Nit: asserting the existence of the endTimestamp might be a little more robust.

src/arns.lua
```diff
@@ -48,18 +50,17 @@ function arns.buyRecord(name, purchaseType, years, from, timestamp, processId)
 		type = purchaseType,
 		undernameLimit = constants.DEFAULT_UNDERNAME_COUNT,
 		purchasePrice = totalRegistrationFee,
+		endTimestamp = purchaseType == "lease" and timestamp + constants.oneYearMs * years or nil,
```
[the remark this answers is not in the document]
arielmelendez: (I realize it's just a ternary but I wish they'd just use the same control characters as everyone else)

src/arns.lua
```diff
@@ -423,14 +424,106 @@ function arns.assertValidIncreaseUndername(record, qty, currentTimestamp)
 	return true
 end
 
+-- AUCTIONS
+
+--- Creates an auction for a given name
+--- @param name string The name of the auction
```
arielmelendez, line 430: Nit: "the name being auctioned" otherwise this reads like it's an auction title

src/arns.lua
arielmelendez: >  if we don't update the state manually
Is that the plan?

src/arns.lua
```diff
@@ -423,14 +424,106 @@ function arns.assertValidIncreaseUndername(record, qty, currentTimestamp)
...
+end
+
+function arns.submitAuctionBid(name, bidAmount, bidder, timestamp, processId, type, years)
+	local auction = arns.getAuction(name)
+	if not auction then
+		error("Auction does not exist")
+	end
+
+	-- assert the bid is between auction start and end timestamps
+	if timestamp < auction.startTimestamp or timestamp > auction.endTimestamp then
```
arielmelendez, line 466: Should endTimestamp be inclusive?

src/arns.lua
```diff
@@ -423,14 +424,106 @@ function arns.assertValidIncreaseUndername(record, qty, currentTimestamp)
...
+		-- TODO: we should likely clean up the auction if it is outside of the time range
```
arielmelendez, line 467: Prune should catch that IFF the endTimestamp is exclusive (i.e. prune would end an auction AT the end timestamp if they overlapped).

src/arns.lua
```diff
@@ -423,14 +424,106 @@ function arns.assertValidIncreaseUndername(record, qty, currentTimestamp)
...
+		error("Bid timestamp is outside of auction start and end timestamps")
+	end
+	local requiredBid = auction:getPriceForAuctionAtTimestamp(timestamp, type, years)
+	local requiredOrBidAmount = bidAmount or requiredBid
```
arielmelendez, line 471: What's the objective here? Let callers NOT supply a bidAmount?

src/arns.lua
```diff
@@ -423,14 +424,106 @@ function arns.assertValidIncreaseUndername(record, qty, currentTimestamp)
...
+	-- check the balance of the bidder
+	if not utils.walletHasSufficientBalance(bidder, finalBidAmount) then
+		error("Insufficient balance")
+	end
+
+	local record = {
+		processId = processId,
+		startTimestamp = timestamp,
+		endTimestamp = type == "lease" and timestamp + constants.oneYearMs * years or nil,
+		undernameLimit = constants.DEFAULT_UNDERNAME_COUNT,
```
arielmelendez, line 487: If the record had previously increased its undername limit, is that lost in transition here?

src/arns.lua
```diff
@@ -423,14 +424,106 @@ function arns.assertValidIncreaseUndername(record, qty, currentTimestamp)
...
+	}
+	local rewardForInitiator = math.floor(finalBidAmount * 0.5)
+	local rewardForProtocol = finalBidAmount - rewardForInitiator
+	-- reduce bidder balance by the final bid amount
+	balances.transfer(auction.initiator, bidder, rewardForInitiator)
+	balances.transfer(ao.id, bidder, rewardForProtocol)
+	arns.removeAuction(name)
+	arns.addRecord(name, record)
+	-- make sure we tally name purchase given, even though only half goes to protocol
+	-- TODO: DO WE WANT TO TALLY THE ENTIRE AMOUNT OR JUST THE REWARD FOR THE PROTOCOL?
```
arielmelendez, line 499: Entire amount IMO

src/arns.lua
```diff
@@ -445,14 +538,26 @@ function arns.pruneRecords(currentTimestamp)
 	return prunedRecords
 end
 
+-- prune auctions that have expired
+function arns.pruneAuctions(currentTimestamp)
+	local prunedAuctions = {}
+	for name, auction in pairs(arns.getAuctions()) do
+		if auction.endTimestamp <= currentTimestamp then
```
arielmelendez, line 545: Note here to make sure that end timestamp is handled consistently in various functions (inclusively or exclusively during bidding, pruning, etc.)

src/auctions.lua
```diff
@@ -0,0 +1,91 @@
+local Auction = {}
+
+-- Default Auction Settings
+AuctionSettings = {
+	durationMs = 60 * 1000 * 60 * 24 * 14, -- 14 days in milliseconds
+	decayRate = 0.000000000016847809193121693, -- 0.02037911 / durationMs
```
arielmelendez, line 6: Note that this is 27 digits after the decimal so keep that in mind if passing across process boundaries.

src/auctions.lua
```diff
@@ -0,0 +1,91 @@
...
+		demandFactor = demandFactor,
+		settings = {
+			durationMs = AuctionSettings.durationMs,
+			decayRate = AuctionSettings.decayRate,
+			scalingExponent = AuctionSettings.scalingExponent,
+			startPriceMultiplier = AuctionSettings.startPriceMultiplier,
+		},
+	}
+	setmetatable(auction, self)
+	self.__index = self
```
arielmelendez, line 47: Seems there's some nuances to setting up the class this way (in the constructor) vs. separately. Will share chatGPT thread with you separately about it.

src/auctions.lua
```diff
@@ -0,0 +1,91 @@
...
+	return auction
+end
+
+--- Computes the prices for the auction.
+--- @param type string The type of auction
+--- @param years number The number of years for calculation
+--- @param intervalMs number The interval in milliseconds, must be at least 15 minutes
```
arielmelendez, line 54: Note to double check where the 15 minutes is enforced.

src/main.lua
```diff
@@ -676,7 +679,14 @@ addEventingHandler(ActionMap.BuyRecord, utils.hasMatchingTag("Action", ActionMap
 	ao.send({
 		Target = msg.From,
 		Tags = { Action = "Buy-Record-Notice", Name = msg.Tags.Name },
-		Data = json.encode(record),
+		Data = json.encode({
```
arielmelendez, line 682: What was the material difference here?

src/main.lua
```diff
@@ -2133,4 +2141,252 @@ end)
 
 -- END READ HANDLERS
 
+-- AUCTION HANDLER
+addEventingHandler("releaseName", utils.hasMatchingTag("Action", ActionMap.ReleaseName), function(msg)
+	-- validate the name and process id exist, then create the auction using the auction function
+	local name = msg.Tags.Name
+	local processId = msg.From
+	local record = arns.getRecord(name)
+	local initiator = msg.Tags.Initiator
```
arielmelendez, line 2150: Can we be getting this from a Forwarded-For tag or similar that's provided by the MU during message forwarding?

src/main.lua
```diff
@@ -2133,4 +2141,252 @@ end)
...
+			Target = msg.From,
+			Action = "Invalid-" .. ActionMap.ReleaseName .. "-Notice",
+			Error = "Invalid-Record-Type",
+			-- only permabought names can be released by the process that owns the name
+		})
+		return
+	end
+
+	-- we should be able to create the auction here
+	local status, auctionOrError = pcall(arns.createAuction, name, tonumber(msg.Timestamp), initiator)
```
arielmelendez, line 2182: It's pretty much free to do these as `eventingPcall`s now since we wrapped all the handlers.

src/main.lua
```diff
@@ -2133,4 +2141,252 @@ end)
...
+		ao.send({
+			Target = msg.From,
+			Action = "Invalid-" .. ActionMap.ReleaseName .. "-Notice",
+			Error = "Auction-Creation-Error",
+			Data = tostring(auctionOrError),
+		})
+		return
+	end
+
+	local auction = {
```
arielmelendez, line 2203: Why is this step necessary?

src/main.lua
```diff
@@ -2133,4 +2141,252 @@ end)
...
+end)
+
+-- Handler to get auction prices for a name
+addEventingHandler("auctionPrices", utils.hasMatchingTag("Action", ActionMap.AuctionPrices), function(msg)
+	local name = string.lower(msg.Tags.Name)
+	local auction = arns.getAuction(name)
+	local timestamp = tonumber(msg.Tags.Timestamp or msg.Timestamp)
+	local type = msg.Tags.Type or "permabuy"
+	local years = msg.Tags.Years and tonumber(msg.Tags.Years) or nil
+	local intervalMs = msg.Tags["Price-Interval-Ms"] and tonumber(msg.Tags["Price-Interval-Ms"]) or 15 * 60 * 1000 -- 15 minute intervals by default
```
arielmelendez, line 2259: We should enforce a minimum interval.

src/main.lua
```diff
@@ -2133,4 +2141,252 @@ end)
...
+			name = auction.name,
+			type = type,
+			years = years,
+			prices = jsonPrices,
+			currentPrice = currentPrice,
+		}),
+	})
+end)
+
+addEventingHandler("auctionBid", utils.hasMatchingTag("Action", ActionMap.AuctionBid), function(msg)
```
arielmelendez, line 2290: Why did we move away from Buy-Record as the one true path for acquiring names?

PR #97, opened 2024-10-15
feat(PE-6949): add delegate instant withdraws

vilenarios:


src/gar.lua
```diff
@@ -361,18 +365,40 @@ function gar.decreaseDelegateStake(gatewayAddress, delegator, qty, currentTimest
...
 
-	-- Lock the qty in a vault to be unlocked after withdrawal period and decrease the gateway's total delegated stake
-	gateway.delegates[delegator].vaults[messageId] = newDelegateVault
-	gateway.delegates[delegator].delegatedStake = gateway.delegates[delegator].delegatedStake - qty
-	gateway.totalDelegatedStake = gateway.totalDelegatedStake - qty
+		-- Add penalty to AR.IO protocol balance
+		balances.increaseBalance(ao.id, penaltyAmount)
+
+		-- Withdraw the remaining tokens to the delegate
+		balances.increaseBalance(delegator, amountToWithdraw)
```
dtfiedler, line 383: shouldn't this be `decreaseBalance`?

vilenarios: Nope.  The user had staked tokens, so we "unlock them", add the penalty amount to the protocol balance, and the rest to the user, and then delete the vault.

dtfiedler: ah, right

src/gar.lua
```diff
@@ -369,18 +369,40 @@ function gar.decreaseDelegateStake(gatewayAddress, delegator, qty, currentTimest
...
+		gateway.delegates[delegator].delegatedStake = gateway.delegates[delegator].delegatedStake - qty
+		gateway.totalDelegatedStake = gateway.totalDelegatedStake - qty
 
-	local newDelegateVault = {
-		balance = qty,
-		startTimestamp = currentTimestamp,
-		endTimestamp = currentTimestamp + gar.getSettings().delegates.withdrawLengthMs,
-	}
+		-- Calculate the penalty amount
+		local maxPenalty = 0.80
```
dtfiedler, line 379: nit: use the constant variable you created

src/gar.lua
```diff
@@ -756,4 +778,61 @@ function gar.cancelDelegateWithdrawal(from, gatewayAddress, vaultId)
...
+
+	-- Ensure the elapsed time is not negative
+	if elapsedTime < 0 then
+		error("Invalid elapsed time")
+	end
+
+	-- Calculate the penalty rate based on elapsed time
+	local maxPenalty = 0.80
+	local minPenalty = 0.05
+	local penaltyRate = maxPenalty - ((maxPenalty - minPenalty) * (elapsedTime / totalWithdrawalTime))
```
dtfiedler, line 809: nit: use the constants you added

src/main.lua
```diff
@@ -1101,6 +1102,68 @@ addEventingHandler(
 	end
 )
 
+addEventingHandler(
```
dtfiedler, line 1105: ah, I see it now - i don't think we need a separate handler for instant - so long as we accept the tag on existing decrease-stake handlers (operators and delegates) and the logic matches

src/main.lua
```diff
@@ -1146,7 +1213,9 @@ addEventingHandler(
 			msg.ioEvent:addField("PreviousStake", newStake + quantity)
 			msg.ioEvent:addField("NewStake", newStake)
 			msg.ioEvent:addField("GatewayTotalDelegatedStake", gateway.totalDelegatedStake)
-
+			if msg.Tags.Instant == true then
+				msg.ioEvent:addField("InstantWithdrawal", true)
+			end
```
dtfiedler, line 1218:
```suggestion
			msg.ioEvent:addField("InstantWithdrawal", insantWithdraw)
```

src/main.lua
```diff
@@ -1127,6 +1193,7 @@ addEventingHandler(
 		local from = utils.formatAddress(msg.From)
 		local target = utils.formatAddress(msg.Tags.Target or msg.Tags.Address)
 		local quantity = tonumber(msg.Tags.Quantity)
+		local instantWithdraw = msg.Tags.Instant == true
```
dtfiedler, line 1196:
```suggestion
		local instantWithdraw = msg.Tags.Instant && msg.Tags.Instant == "true"
```

src/gar.lua
vilenarios: https://github.com/ar-io/ar-io-network-process/pull/97/commits/0f8b37a3d29046fead347d62fe2cf4f9f146c711

src/gar.lua
vilenarios: Done.

https://github.com/ar-io/ar-io-network-process/pull/97/commits/0f8b37a3d29046fead347d62fe2cf4f9f146c711

src/main.lua
```diff
@@ -81,6 +81,7 @@ local ActionMap = {
 	DelegateStake = "Delegate-Stake",
 	DecreaseDelegateStake = "Decrease-Delegate-Stake",
 	CancelDelegateWithdrawal = "Cancel-Delegate-Withdrawal",
+	InstantDelegateWithdrawal = "Instant-Delegate-Withdrawal",
```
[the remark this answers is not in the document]
vilenarios: We have this already.

InstantDelegateWithdrawal is what a delegate would run to instantly unlock tokens that were already in a vault.  It accepts a gateway and a vault ID.

DecreaseDelegateStake with the `instant=true` tag would instantly unlock tokens for the max penalty.

If we were to combine DecreaseDelegateStake with InsteadDelegateWithdrawl, we need to also pass a vault ID.  You think its worthwhile to still combine these even though they have different impacts?  We split out CancelDelegateWithdrawal instead of adding that to DecreaseDelegateStake, which is what I modeled InstantDelegateWIthdrawal after.

src/main.lua
vilenarios: So should we add Vault-ID in here too?  I could still combine the logic in some utility functions but maintain separate handlers.

src/main.lua
vilenarios: Double && is not valid in lua.  Fixed this

https://github.com/ar-io/ar-io-network-process/pull/97/commits/acdb40836eb10c0afa4556b80f05226df37ddb84

src/main.lua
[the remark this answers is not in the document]
dtfiedler: ah i see - thanks for the clarification. in that case, it makes sense to leave them as separate

src/main.lua
dtfiedler: [a0c17c3](https://github.com/ar-io/ar-io-network-process/pull/97/commits/a0c17c3c061c235b3e4565caeb92ee4ab155ea8c)

dtfiedler (approved):

PR #98, opened 2024-10-15
fix(gar): make sure we revert balances on pruning if failure, update monitor tests

dtfiedler:
This caused the protocol balance to get slashes rewards multiple times when ticking should have failed and reverted the balances


src/main.lua
```diff
@@ -151,12 +151,13 @@ end, function(msg)
 	local msgId = msg.Id
 	local msgTimestamp = tonumber(msg.Timestamp)
 	print("Pruning state at timestamp: " .. msgTimestamp)
+	-- we may to be concious about deep copying here, as it could consume a large amount of memory. so long as we are pruning effectively, this should be fine
```
arielmelendez, line 154: nit: typo

PR #102, opened 2024-10-17
fix(vaults): fix `vaults`, including `increaseVault`, `extendVault`, `vaultedTransfer`, and add tests for coverage

dtfiedler:


src/main.lua
```diff
@@ -386,47 +383,53 @@ addEventingHandler(ActionMap.VaultedTransfer, utils.hasMatchingTag("Action", Act
...
+		-- sender gets an immediate credit notice
 		ao.send({
-			Target = msg.From,
-			Recipient = msg.Tags.Recipient,
-			Quantity = msg.Tags.Quantity,
-			Tags = { Action = "Debit-Notice" },
+			Target = from,
+			Recipient = recipient,
+			Quantity = quantity,
+			Tags = { Action = "Credit-Notice", ["Vault-Id"] = msgId },
```
arielmelendez, line 407: Why did we move away from a Debit-Notice here?

arielmelendez (approved):

src/main.lua
dtfiedler: bc i'm a fool and got the definitions mixed up in my head - reverted back in [7d48ad4](https://github.com/ar-io/ar-io-network-process/pull/102/commits/7d48ad4516cf276782179f9b0f8805de4362ec1a)

PR #104, opened 2024-10-17
feat(events): instrument token supply handler PE-6976

arielmelendez:


src/main.lua
```diff
@@ -1335,41 +1343,59 @@ addEventingHandler("totalTokenSupply", utils.hasMatchingTag("Action", "Total-Tok
...
+	lastKnownDelegatedSupply = delegatedSupply
+	lastKnownWithdrawSupply = withdrawSupply
+
+	msg.ioEvent:addField("Total-Token-Supply", lastKnownTotalSupply)
+	msg.ioEvent:addField("Circulating-Supply", lastKnownCirculatingSupply)
+	msg.ioEvent:addField("Locked-Supply", lastKnownLockedSupply)
+	msg.ioEvent:addField("Staked-Supply", lastKnownStakedSupply)
+	msg.ioEvent:addField("Delegated-Supply", lastKnownDelegatedSupply)
+	msg.ioEvent:addField("Withdraw-Supply", lastKnownWithdrawSupply)
+	msg.ioEvent:addField("Protocol-Balance", protocolBalance)
```
dtfiedler, line 1397: FWIW - this is mostly used as a dryRun only handler - so it's unlikely to ever surface event data. that being said, we can move the computation to a separate function and call it before/after every write message and add the event data you've added here

arielmelendez: I was toying with factoring this out and calling it any time functions that affect the supplies get called but I worry that it could get unnecessarily computationally expensive at scale (also from a gas perspective). I haven't ruled it out for ticks so that we get it all at least once a day.

PR #107, opened 2024-10-18
feat(events): track supply changes more closely PE-6796

arielmelendez:


src/main.lua
```diff
@@ -809,7 +870,7 @@ addEventingHandler(ActionMap.JoinNetwork, utils.hasMatchingTag("Action", ActionM
 	local timestamp = tonumber(msg.Timestamp)
 
 	msg.ioEvent:addField("Resolved-Observer-Address", formattedObserverAddress)
-	msg.ioEvent:addField("Sender-Previous-Balance", fromBalance)
+	msg.ioEvent:addField("Sender-Previous-Balance", Balances[fromAddress] or 0)
```
dtfiedler, line 873: nit - 
```suggestion
	msg.ioEvent:addField("Sender-Previous-Balance", balances.getBalance(fromAddress))
```

src/main.lua
```diff
@@ -822,14 +883,19 @@ addEventingHandler(ActionMap.JoinNetwork, utils.hasMatchingTag("Action", ActionM
 		return
 	end
 
-	msg.ioEvent:addField("Sender-New-Balance", fromBalance)
+	msg.ioEvent:addField("Sender-New-Balance", Balances[fromAddress] or 0)
```
dtfiedler, line 886:
```suggestion
	msg.ioEvent:addField("Sender-New-Balance", balances.getBalance(fromAddress))
```

src/vaults.lua
```diff
@@ -139,18 +139,22 @@ end
...
 				balances.increaseBalance(owner, nestedVault.balance)
-				vaults[id] = nil
+				ownersVaults[id] = nil
+				prunedVaults[id] = nestedVault
 			end
 		end
 		-- update the owner vault
-		allVaults[owner] = vaults
+		-- TODO: I THINK THIS LINE IS UNNECESSARY. CHECK TESTS
+		allVaults[owner] = ownersVaults
```
dtfiedler, line 153: it's not - as we are updating the objects in place above, so not necessary - this was likely leftover from when we ported it over from smartweave

src/main.lua
arielmelendez: My issue with that approach is it makes a whole deep copy of the balances table to then just throw it away. Why do that here for a scalar value?

src/vaults.lua
arielmelendez: Excised and tests ran fine:
be82619

src/main.lua
arielmelendez: Perhaps what we need is a cheaper getter function if you want to protect accesses of the main Balances object?

src/main.lua
```diff
@@ -882,6 +955,11 @@ addEventingHandler(ActionMap.LeaveNetwork, utils.hasMatchingTag("Action", Action
 	msg.ioEvent:addField("Joined-Gateways-Count", gwStats.joined)
 	msg.ioEvent:addField("Leaving-Gateways-Count", gwStats.leaving)
 
+	lastKnownStakedSupply = lastKnownStakedSupply - gwPrevStake - gwPrevTotalDelegatedStake
+	lastKnownWithdrawSupply = lastKnownWithdrawSupply + gwPrevStake + gwPrevTotalDelegatedStake
+	msg.ioEvent:addField("Staked-Supply", lastKnownStakedSupply)
+	msg.ioEvent:addField("Withdraw-Supply", lastKnownWithdrawSupply)
```
arielmelendez, line 961: TODO: Add event payload data for gw's remaining totalDelegatedStake which should damn well be 0!

src/gar.lua
```diff
@@ -654,6 +655,9 @@ function gar.pruneGateways(currentTimestamp, msgId)
 	local result = {
 		prunedGateways = {},
 		slashedGateways = {},
+		gatewayStakeReturned = 0,
+		delegateStakeReturned = 0,
+		stakeSlashed = 0,
```
arielmelendez, line 660: TODO: Add pruned gws with non-zero totalDelegatedStake to use in event data.

src/main.lua
```diff
@@ -197,28 +197,60 @@ end, function(msg)
...
 		end
 	end
 
+	msg.ioEvent:addField(
+		"Total-Supply",
+		lastKnownCirculatingSupply
+			+ lastKnownLockedSupply
+			+ lastKnownStakedSupply
+			+ lastKnownDelegatedSupply
+			+ lastKnownWithdrawSupply
```
arielmelendez, line 245: Todo: add protocol balance

src/main.lua
```diff
@@ -197,28 +197,60 @@ end, function(msg)
...
+		local slashedGatewaysCount = utils.lengthOfTable(slashedGateways or {})
 		if slashedGatewaysCount > 0 then
-			msg.ioEvent:addField("Slashed-Gateways", resultOrError.slashedGateways)
+			msg.ioEvent:addField("Slashed-Gateways", slashedGateways)
 			msg.ioEvent:addField("Slashed-Gateways-Count", slashedGatewaysCount)
 		end
 	end
 
+	msg.ioEvent:addField(
+		"Total-Supply",
```
arielmelendez, line 240: Update to Total-Token-Supply

src/main.lua
```diff
@@ -197,28 +197,60 @@ end, function(msg)
...
+		lastKnownCirculatingSupply
+			+ lastKnownLockedSupply
+			+ lastKnownStakedSupply
+			+ lastKnownDelegatedSupply
+			+ lastKnownWithdrawSupply
+	)
+	msg.ioEvent:addField("Circulating-Supply", lastKnownCirculatingSupply)
```
arielmelendez, line 247: Consider only setting these here if there is a change in value.

src/main.lua
arielmelendez: 6b1688e

src/main.lua
arielmelendez: 6b1688e

src/main.lua
arielmelendez: 6b1688e

src/main.lua
arielmelendez: cfec28d

src/gar.lua
arielmelendez: cfec28d

PR #124, opened 2024-10-23
fix(auctions): small fix for auctions, add handlers

dtfiedler:


src/main.lua
```diff
@@ -2256,7 +2285,7 @@ addEventingHandler("auctionPrices", utils.hasMatchingTag("Action", ActionMap.Auc
 	local name = string.lower(msg.Tags.Name)
 	local auction = arns.getAuction(name)
 	local timestamp = tonumber(msg.Tags.Timestamp or msg.Timestamp)
-	local type = msg.Tags.Type or "permabuy"
+	local type = msg.Tags["Purchase-Type"] or "permabuy"
```
arielmelendez, line 2288: Should there be an analogous change in a test somewhere?

arielmelendez (approved):

src/main.lua
dtfiedler: the current integration test validates the defaults. will add another for providing one

dtfiedler: i'll add a test with different tags for lease expiration ones to validate these

PR #126, opened 2024-10-23
feat(auctions): introduce lease expiration initiated auctions

dtfiedler:
Instead of pruning records, the names will go into auction with the rewards all given to the protocol balance


arielmelendez (dismissed):

tools/constants.mjs
```diff
@@ -10,8 +10,9 @@ export const PROCESS_OWNER = ''.padEnd(43, '1');
 export const STUB_ADDRESS = ''.padEnd(43, '2');
 export const INITIAL_PROTOCOL_BALANCE = 50_000_000_000_000; // 50M IO
 export const INITIAL_OWNER_BALANCE = 950_000_000_000_000; // 950M IO
-export const STUB_TIMESTAMP = 1704088800000; // 01-01-2024 00:00:00
+export const STUB_TIMESTAMP = 21600000; // 01-01-1970 00:00:00
```
dtfiedler, line 13: this is before any epochs - we'll manually set timestamps where we want to test epoch creation/distributions

PR #127, opened 2024-10-24
fix(gar): prune gateways immediately once they hit 30 consecutive fai…

dtfiedler:
…lures

### DIFF
--- a/src/main.lua
+++ b/src/main.lua
@@ -256,14 +256,24 @@ end, function(msg)
 		end
 
 		local pruneGatewayResults = resultOrError.pruneGatewayResults or {}
+
 		lastKnownCirculatingSupply = lastKnownCirculatingSupply
 			+ (pruneGatewayResults.delegateStakeReturned or 0)
 			+ (pruneGatewayResults.gatewayStakeReturned or 0)
+
 		lastKnownWithdrawSupply = lastKnownWithdrawSupply
 			- (pruneGatewayResults.delegateStakeReturned or 0)
 			- (pruneGatewayResults.gatewayStakeReturned or 0)
+			+ (pruneGatewayResults.delegateStakeWithdrawing or 0)
+			+ (pruneGatewayResults.gatewayStakeWithdrawing or 0)
+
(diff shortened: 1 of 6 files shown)


src/main.lua
```diff
@@ -1754,6 +1762,10 @@ addEventingHandler("distribute", utils.hasMatchingTag("Action", "Tick"), functio
 	if #newDemandFactors > 0 then
 		msg.ioEvent:addField("New-Demand-Factors", newDemandFactors, ";")
 	end
+	if #newPrunedGateways > 0 then
```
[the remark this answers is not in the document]
dtfiedler: we'll actually need to redo all of that here - as it has the same potential impacts for supply 😢

src/main.lua
```diff
@@ -1655,18 +1655,22 @@ addEventingHandler("distribute", utils.hasMatchingTag("Action", "Tick"), functio
...
 		local distributedEpoch = epochs.distributeRewardsForEpoch(timestamp)
 		if distributedEpoch ~= nil and distributedEpoch.epochIndex ~= nil then
 			tickedRewardDistributions[tostring(distributedEpoch.epochIndex)] =
 				distributedEpoch.distributions.totalDistributedRewards
 			totalTickedRewardsDistributed = totalTickedRewardsDistributed
 				+ distributedEpoch.distributions.totalDistributedRewards
 		end
-
+		-- prune any gateway that has hit the failed 30 consecutive epoch threshold after the epoch has been distributed
+		local prunedGateways = gar.pruneGateways(timestamp)
```
arielmelendez, line 1667: This returns more than just gateways:
image

dtfiedler: ah - right, here we would actually only care about `slashedGateways` and `stakeSlashed`. no delegate stake would be returned (as it was already computed and done by the `prune` that is called before this).

conversely, in `prune` we should not need to account for `stakeSlashed` or `slashedGateways` as that should will happen on this handler, not in the `prune`

arielmelendez (changes requested):
I'm going to push a commit to properly track the circulating supply changes here.

arielmelendez (approved):

PR #129, opened 2024-10-25
PE-6976: instant withdraw event data

arielmelendez:
- Corrected spelling of `expendited` -> `expedited` everywhere necessary
- Updated constant names from `FEE` to `RATE` where sensible
- Modified redundant unit test for instant delegate withdrawals to the max duration (non-invariant boundary case)
- Collected penalty rate and fee information for event data on instant withdrawals


vilenarios (approved):
Lookin good!  More refactors coming in the instant gateway withdraw branch.

PR #132, opened 2024-10-25
PE-7033 slashing history

arielmelendez:


spec/gar_spec.lua
```diff
@@ -1743,17 +1743,23 @@ describe("gar", function()
...
 			_G.GatewayRegistry[stubGatewayAddress] = {
 				operatorStake = gar.getSettings().operators.minStake,
-				totalDelegatedStake = 0,
+				totalDelegatedStake = 123,
 				vaults = {},
 				delegates = {},
 			}
-			local status, err = pcall(gar.slashOperatorStake, stubGatewayAddress, slashAmount)
+			local currentTimestamp = 123456
+			local status, err = pcall(gar.slashOperatorStake, stubGatewayAddress, slashAmount, currentTimestamp)
```
dtfiedler, line 1751: in these tests, we can stop using `pcall` - we want them to work as expected and the pcall only wraps any errors and makes the test failure ambigous. said another way - we should shift to using pcall _only_ when we want to validate the error of a given unit test

dtfiedler (approved):
small nit in the test, but otherwise looks good

PR #133, opened 2024-10-26
feat(PE-6980): all gateways to use instant withdraw for a fee

vilenarios:
Adds support for gateway operators to instantly withdraw their stake or immediately unlock a vault for an early withdrawal fee.


src/gar.lua
```diff
@@ -855,4 +870,48 @@ function gar.instantDelegateWithdrawal(from, gatewayAddress, vaultId, currentTim
...
+	local elapsedTime = currentTimestamp - vault.startTimestamp
+	local totalWithdrawalTime = gar.getSettings().operators.withdrawLengthMs
+
+	-- Ensure the elapsed time is not negative
+	if elapsedTime < 0 then
+		error("Invalid elapsed time")
+	end
+
+	-- Process the instant withdrawal
+	local expeditedWithdrawalFee, amountToWithdraw =
```
arielmelendez, line 899: Need to collect the penaltyRate here to be used down below.

src/gar.lua
```diff
@@ -855,4 +870,48 @@ function gar.instantDelegateWithdrawal(from, gatewayAddress, vaultId, currentTim
...
+		error("Gateway not found")
+	end
+
+	local vault = gateway.vaults[vaultId]
+	if vault == nil then
+		error("Vault not found")
+	end
+
+	if vaultId == from then
+		error("This gateway is leaving and this vault cannot be instantly withdrawn.")
```
arielmelendez, line 886: Was this to spec or did we decide that a leaving gateway can eat the penalties too?

src/gar.lua
```diff
@@ -185,15 +209,26 @@ function gar.decreaseOperatorStake(from, qty, currentTimestamp, msgId)
...
-		balance = qty,
-		startTimestamp = currentTimestamp,
-		endTimestamp = currentTimestamp + gar.getSettings().operators.withdrawLengthMs,
-	}
-	-- update the gateway
+
+	-- Update the gateway
 	GatewayRegistry[from] = gateway
+
 	return gar.getGateway(from)
```
arielmelendez, line 232: Should be pretty straightforward to update this to return a table containing the gateway and the penalty info that we want to event on later.

src/gar.lua
vilenarios: Yes - a leaving gateway can eat the penalties only on their stake aove the minimum.  The minimum stake vault is not allowed to be instantly withdrawn, and they must wait the full 90 days.

src/gar.lua
vilenarios: Doh... fixed.

https://github.com/ar-io/ar-io-network-process/commit/16d793cf9182e98d8b3a316030a48f97c3fc6242

src/gar.lua
vilenarios: Updated decreaseOperatorStake and decreaseDelegateStake

https://github.com/ar-io/ar-io-network-process/commit/1e636ccb7984857e9dde09385f64171752a39d75

src/gar.lua
```diff
@@ -173,7 +197,7 @@ function gar.decreaseOperatorStake(from, qty, currentTimestamp, msgId)
 	end
 
 	if gateway.status == "leaving" then
-		error("Gateway is leaving the network and withdraw more stake.")
+		error("Gateway is leaving the network and cannot withdraw more stake.")
```
arielmelendez, line 200: 👌

vilenarios: I found another error message that was incorrectly worded like this one in decreaseDelegateStake.  Fixed it!

src/main.lua
```diff
@@ -1151,8 +1172,83 @@ addEventingHandler(
 
 		ao.send({
 			Target = msg.From,
-			Tags = { Action = "Decrease-Operator-Stake-Notice" },
-			Data = json.encode(gateway),
+			Tags = {
+				Action = "Decrease-Operator-Stake-Notice",
+				["Penalty-Rate"] = tostring(decreaseOperatorStakeResult.penaltyRate),
```
arielmelendez, line 1177: Is it the case that we HAVE to tostring all our numbers that we pass to Tags?

src/main.lua
```diff
@@ -1151,8 +1172,83 @@ addEventingHandler(
...
+		end
+
+		addSupplyData(msg.ioEvent)
+
+		ao.send({
+			Target = msg.From,
+			Tags = {
+				Action = "Instant-Operator-Withdrawal-Notice",
+				["Vault-Id"] = msg.Tags["Vault-Id"],
+				["Amount-Withdrawn"] = amountWithdrawn,
```
arielmelendez, line 1249: Re: tags - do we need a tostring for our currency numbers?

arielmelendez (approved):

arielmelendez (approved):

PR #146, opened 2024-10-28
fix(gar): autoStake all delegate rewards

dtfiedler:
All delegates will have their rewards restaked against the rewarded gateway, rather than going to their balance. This is default behavior and cannot be overwritten

### DIFF
--- a/spec/epochs_spec.lua
+++ b/spec/epochs_spec.lua
@@ -684,15 +684,14 @@ describe("epochs", function()
 			}
 
 			local epochIndex = 0
-			local originalDelegatedStake = 100000000
 			for i = 1, 5 do
 				local gateway = {
 					operatorStake = gar.getSettings().operators.minStake,
-					totalDelegatedStake = originalDelegatedStake,
+					totalDelegatedStake = 100000000,
 					vaults = {},
 					delegates = {
 						["this-is-a-delegate"] = {
-							delegatedStake = originalDelegatedStake,
+							delegatedStake = 100000000,
 							startTimestamp = 0,
 							vaults = {},
 						},
@@ -714,7 +713,7 @@ describe("epochs", function()
 						port = 443,
 						allowDelegatedStaking = true,
 						minDelegatedStake = 100,
-						autoStake = i == 2, -- set autostake on for the second gateway
+						autoStake = i ~= 5, -- set autostake on for all gateways except the last one
 						label = "test",
 						properties = "",
 						delegateRewardShareRatio = 10,
@@ -731,7 +730,7 @@ describe("epochs", function()
 			local expectedObserverReward = epoch.distributions.totalEligibleObserverReward
 
 			-- distribute rewards for the epoch
-			local status, result = pcall(epochs.distributeRewardsForEpoch, epoch.distributionTimestamp)
+			local status = pcall(epochs.distributeRewardsForEpoch, epoch.distributionTimestamp)
 			assert.is_true(status)
 			-- gateway 1 should not get any rewards - failed observation and did not observe, should not get any rewards
 			assert.are.same({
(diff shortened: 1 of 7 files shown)


spec/epochs_spec.lua
```diff
@@ -684,14 +684,15 @@ describe("epochs", function()
 			}
 
 			local epochIndex = 0
+			local originalDelegatedStake = 100000000
 			for i = 1, 5 do
 				local gateway = {
 					operatorStake = gar.getSettings().operators.minStake,
-					totalDelegatedStake = 100000000,
+					totalDelegatedStake = originalDelegatedStake,
```
arielmelendez, line 691: As a reader and test debugger this past week, I've found the use of variables in tests for testing values to be a detractor from readability, SEO, and debuggability. In the past I've found small comments accompanying hardcoded values to provide context as a useful way to go instead.

Using variables in this way starts to lean towards writing too much logic into TESTING code that itself could be faulty or hide problems when conditions dynamically change in the surrounding test code. I prefer the security, SEO, and debuggability provided by rigid constants.

src/epochs.lua
```diff
@@ -586,7 +586,9 @@ function epochs.distributeRewardsForEpoch(currentTimestamp)
 					local actualDelegateReward = math.floor(eligibleDelegateReward * percentOfEligibleEarned)
 					-- distribute the rewards to the delegate if greater than 0
 					if actualDelegateReward > 0 then
-						balances.transfer(delegateAddress, ao.id, actualDelegateReward)
+						-- increase the stake and decrease the protocol balance
+						gar.increaseExistingDelegateStake(gatewayAddress, delegateAddress, actualDelegateReward)
```
arielmelendez, line 590: Every call to this is going to create a copy of the gateway object. Might want to factor out fetching the gateway object outside the loop and then pass the gw into the function to save on wasteful compute.

arielmelendez: Might be worth a search around the contract for patterns like that too... i.e. wasteful copying in loops.

arielmelendez (changes requested):
See my comment re: wasteful copying.

spec/epochs_spec.lua
dtfiedler: fair - this one is used 7 times in this test align to compute expected rewards, but will adjust

src/epochs.lua
dtfiedler: refer to the changes here - https://github.com/ar-io/ar-io-network-process/pull/146/commits/dcaa064d7d3d28cbac25e0909b5c8e0e5e6dc739

reduces duplicate copies within this handler - the call signature comments help with static analysis

src/balances.lua
```diff
@@ -20,15 +25,24 @@ function balances.transfer(recipient, from, qty)
...
+	local balances = utils.deepCopy(Balances) or {}
+	return balances
 end
 
+--- Reduces the balance of an address
+---@param target string The address to reduce balance for
+---@param qty number The amount to reduce by (must be integer)
+---@throws error If target has insufficient balance
 function balances.reduceBalance(target, qty)
 	local prevBalance = balances.getBalance(target)
```
arielmelendez, line 47: This is likely another unnecessary copy of ALL the balances just to update one value in place.

src/balances.lua
```diff
@@ -38,12 +52,21 @@ function balances.reduceBalance(target, qty)
 	Balances[target] = prevBalance - qty
 end
 
+--- Increases the balance of an address
+---@param target string The address to increase balance for
+---@param qty number The amount to increase by (must be integer)
 function balances.increaseBalance(target, qty)
 	assert(utils.isInteger(qty), debug.traceback("Quantity must be an integer: " .. qty))
 	local prevBalance = balances.getBalance(target) or 0
```
arielmelendez, line 60: Unnecessary copy of all balances to do an in-place balance update below.

src/epochs.lua
arielmelendez: Looks great, thanks!

PR #149, opened 2024-10-28
chore(lint): use luacheck to enforce linting rules, add husky for pre-commit checks

dtfiedler:


src/ao_event.lua
```diff
@@ -89,19 +89,6 @@ local function AOEvent(initialData)
...
-		-- Escape backslashes first
-		s = string.gsub(s, "\\", "\\\\")
-		-- Escape double quotes
-		s = string.gsub(s, '"', '\\"')
-		-- Escape other control characters (optional for full JSON compliance)
-		s = string.gsub(s, "\n", "\\n")
-		s = string.gsub(s, "\r", "\\r")
-		s = string.gsub(s, "\t", "\\t")
-		return s
-	end
```
dtfiedler, line 103: unused function flagged by linting

src/main.lua
```diff
@@ -1164,6 +1164,7 @@ addEventingHandler(
 					{ "balance", "startTimestamp", "endTimestamp" }
 				)
 			end
+			msg.ioEvent:addField("Previous-Operator-Stake", previousStake)
```
arielmelendez, line 1167: Interesting... was moving this here the linter's choice?

arielmelendez (approved):

src/main.lua
dtfiedler: we increment after we initially set the field - and the linter said unused, modified variable

PR #150, opened 2024-10-29
feat(PE-6996): allow upgrading leased names

dtfiedler:
Allows `leased` names to be upgraded to a `permabuy` - can be performed by anyone and is just the cost of permabuying the name. There is no discount applied for the original leasing of the name.

### DIFF
--- a/src/arns.lua
+++ b/src/arns.lua
@@ -23,7 +23,7 @@ function arns.buyRecord(name, purchaseType, years, from, timestamp, processId)
 		years = 1 -- set to 1 year by default
 	end
 
-	local baseRegistrationFee = demand.getFees()[#name]
+	local baseRegistrationFee = demand.baseFeeForNameLength(#name)
 
 	local totalRegistrationFee =
 		arns.calculateRegistrationFee(purchaseType, baseRegistrationFee, years, demand.getDemandFactor())
@@ -95,7 +95,7 @@ function arns.extendLease(from, name, years, currentTimestamp)
 	local record = arns.getRecord(name)
 	-- throw error if invalid
 	arns.assertValidExtendLease(record, currentTimestamp, years)
-	local baseRegistrationFee = demand.getFees()[#name]
+	local baseRegistrationFee = demand.baseFeeForNameLength(#name)
 	local totalExtensionFee = arns.calculateExtensionFee(baseRegistrationFee, years, demand.getDemandFactor())
 
 	if balances.getBalance(from) < totalExtensionFee then
@@ -135,7 +135,7 @@ function arns.increaseundernameLimit(from, name, qty, currentTimestamp)
 		yearsRemaining = arns.calculateYearsBetweenTimestamps(currentTimestamp, record.endTimestamp)
 	end
 
-	local baseRegistrationFee = demand.getFees()[#name]
+	local baseRegistrationFee = demand.baseFeeForNameLength(#name)
 	local additionalUndernameCost =
 		arns.calculateUndernameCost(baseRegistrationFee, qty, record.type, yearsRemaining, demand.getDemandFactor())
 
@@ -304,7 +304,7 @@ function arns.assertValidExtendLease(record, currentTimestamp, years)
 		error("Name is permabought and cannot be extended")
 	end
 
-	if record.endTimestamp and record.endTimestamp + constants.gracePeriodMs < currentTimestamp then
+	if arns.recordExpired(record, currentTimestamp) then
 		error("Name is expired")
 	end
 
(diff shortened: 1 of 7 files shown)


src/arns.lua
```diff
@@ -399,6 +399,48 @@ function arns.getTokenCost(intendedAction)
...
+	local record = arns.getRecord(name)
+	if not record then
+		error("Name is not registered")
+	end
+
+	if record.type == "permabuy" then
+		error("Record is already a permabuy")
+	end
+
+	if record.endTimestamp and record.endTimestamp + constants.gracePeriodMs < currentTimestamp then
```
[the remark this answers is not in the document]
arielmelendez: e.g. `arns.recordIsInGracePeriod(record)`

src/arns.lua
```diff
@@ -399,6 +399,48 @@ function arns.getTokenCost(intendedAction)
...
+		NameRegistry.records[name] = nil
+		error("Name is expired")
+	end
+
+	local baseFee = demand.getFees()[#name]
```
arielmelendez, line 422: Similarly: `demand.baseFeeForNameLength(#name)`

src/arns.lua
```diff
@@ -399,6 +399,48 @@ function arns.getTokenCost(intendedAction)
...
+
+	if not utils.walletHasSufficientBalance(from, upgradeCost) then
+		error("Insufficient balance")
+	end
+
+	record.endTimestamp = nil
+	record.type = "permabuy"
+	record.purchasePrice = upgradeCost
+
+	balances.transfer(ao.id, from, upgradeCost)
```
arielmelendez, line 434: This API always freaks me out when I read it because the flow of the params is `destination, origin, amount` whereas my brain expects `origin, destination, amount`. Can we get any protection from static analysis tools by using a table of args as an input argument along with docblocks?

src/arns.lua
```diff
@@ -399,6 +399,48 @@ function arns.getTokenCost(intendedAction)
...
+	demand.tallyNamePurchase(upgradeCost)
+
+	NameRegistry.records[name] = record
+	return {
```
arielmelendez, line 438: Will be watching for the cost data that gets passed back for event data here.

src/arns.lua
```diff
@@ -399,6 +399,48 @@ function arns.getTokenCost(intendedAction)
...
+	local demandFactor = demand.getDemandFactor()
+	local upgradeCost = arns.calculatePermabuyFee(baseFee, demandFactor)
+
+	if not utils.walletHasSufficientBalance(from, upgradeCost) then
+		error("Insufficient balance")
+	end
+
+	record.endTimestamp = nil
+	record.type = "permabuy"
+	record.purchasePrice = upgradeCost
```
arielmelendez, line 432: I guess we lose the original leasing price information once we make this assignment? I suppose that's fine, but thought I'd raise that in case we have UI data presentation aspirations that include both.

src/main.lua
```diff
@@ -70,7 +70,8 @@ local ActionMap = {
 	VaultedTransfer = "Vaulted-Transfer",
 	ExtendVault = "Extend-Vault",
 	IncreaseVault = "Increase-Vault",
-	BuyRecord = "Buy-Record",
+	BuyRecord = "Buy-Record", -- TODO: standardize these as `Buy-Name` or `Upgrade-Record`
+	UpgradeName = "Upgrade-Name", -- TODO: may be more aligned to `Upgrade-Record`
```
arielmelendez, line 74: Honestly, upgrade is a bit ambiguous. `Permabuy-Record` seems more fitting, but not sure if that creates confusion for users as to whether they should use Buy-Record or Permabuy-Record for an initial (non-lease or "upgrade") purchase.

src/main.lua
```diff
@@ -720,6 +721,58 @@ addEventingHandler(ActionMap.BuyRecord, utils.hasMatchingTag("Action", ActionMap
 	})
 end)
 
+addEventingHandler("upgradeName", utils.hasMatchingTag("Action", ActionMap.UpgradeName), function(msg)
+	-- get the name
+	local name = string.lower(msg.Tags.Name)
```
arielmelendez, line 726: We see these kinds of transformations so often that I wonder if we'd save some code lines by having a prepend handler that just sets these on the message as like:
- normalizedName
- normalizedFrom
- timestampMs
Could also move the creation of the ioEvent there too.

src/main.lua
```diff
@@ -720,6 +721,58 @@ addEventingHandler(ActionMap.BuyRecord, utils.hasMatchingTag("Action", ActionMap
...
+	if not name then
+		ao.send({
+			Target = from,
+			Tags = { Action = "Invalid-" .. ActionMap.UpgradeName .. "-Notice", Error = "Bad-Input" },
+			Data = "Name is required",
+		})
+		return
+	end
+
+	local status, updatedRecordOrError = pcall(arns.upgradeRecord, from, name, timestamp)
```
arielmelendez, line 748: You can use eventingPcall here without consideration for other event data that might be relevant.

arielmelendez (dismissed):
Approved with some suggestions

src/arns.lua
dtfiedler: we can provide it in eventing data - but yes, the state will be lost on upgrade

src/main.lua
dtfiedler: I agree - that's something we can flush out in a separate PR as it'll be useful on every handler

src/arns.lua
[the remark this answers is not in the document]
dtfiedler: [2370035](https://github.com/ar-io/ar-io-network-process/pull/150/commits/23700353209d7919cc841a0344db5113e5e9fd47)

src/arns.lua
dtfiedler: [2370035](https://github.com/ar-io/ar-io-network-process/pull/150/commits/23700353209d7919cc841a0344db5113e5e9fd47)

src/arns.lua
dtfiedler: will save for a separate PR - the added doc blocks do flag, but our current tooling to doesn't support named object params

src/main.lua
dtfiedler: [4730b49](https://github.com/ar-io/ar-io-network-process/pull/150/commits/4730b490caeb77ded5e39908df2e0b3223aeac88)

src/arns.lua
```diff
@@ -395,24 +391,113 @@ function arns.getTokenCost(intendedAction)
...
+	balances.transfer(ao.id, from, upgradeCost)
+	demand.tallyNamePurchase(upgradeCost)
+
+	NameRegistry.records[name] = record
+	return {
+		name = name,
+		record = record,
+		totalUpgradeFee = upgradeCost,
+		baseRegistrationFee = baseFee,
+		remainingBalance = balances.getBalance(from),
```
arielmelendez, line 448: These two aren't as important to return since you can access them from the globals. ndb though

src/arns.lua
```diff
@@ -395,24 +391,113 @@ function arns.getTokenCost(intendedAction)
...
+		protocolBalance = balances.getBalance(ao.id),
+		df = demand.getDemandFactorInfo(),
```
arielmelendez, line 450: nit that this was already fetched above in the `demandFactor` variable.

arielmelendez: OH I see now that it's the full DF stats object and not just the DF value. Gotcha.

arielmelendez (approved):

PR #154, opened 2024-10-30
chore(events): add event data to auctions

dtfiedler:


src/main.lua
```diff
@@ -2695,71 +2725,70 @@ addEventingHandler("auctionBid", utils.hasMatchingTag("Action", ActionMap.Auctio
...
 		return
 	end
 
-	local status, auctionBidOrError =
-		pcall(arns.submitAuctionBid, name, bidAmount, bidder, timestamp, processId, type, years)
-	if not status then
+	local shouldContinue2, result = eventingPcall(msg.ioEvent, function(error)
 		ao.send({
 			Target = msg.From,
-			Action = "Invalid-" .. ActionMap.AuctionBid .. "-Notice",
```
arielmelendez, line 2726: Is this the incorrect way to do this or is there just an advantage to setting these to Tags manually?

src/main.lua
```diff
@@ -138,6 +138,25 @@ local function addRecordResultFields(ioEvent, result)
 	end
 end
 
+local function addAuctionResultFields(ioEvent, result)
+	ioEvent:addFieldsIfExist(result, {
+		"bidAmount",
+		"bidder",
```
arielmelendez, line 144: Will this always be the same as the from field? If so we can likely omit this one.

src/main.lua
```diff
@@ -138,6 +138,25 @@ local function addRecordResultFields(ioEvent, result)
...
+		"rewardForInitiator",
+		"rewardForProtocol",
+		-- TODO: we could return some computed values of the bid relative to start and end price
```
arielmelendez, line 147: Is it hard to get the start and end price values of the auction? I think it would be very valuable data to have here.

arielmelendez (dismissed):

src/main.lua
dtfiedler: we have some incosistency accross notices that this is addressing

src/main.lua
dtfiedler: we compute it based on the bid terms (years, and type) - we need to bubble it up from other places

src/main.lua
dtfiedler: it will, can remove

src/main.lua
dtfiedler: added here - [5931635](https://github.com/ar-io/ar-io-network-process/pull/154/commits/593163517498d51631ceb6c39cd5f54592398589)

PR #157, opened 2024-11-01
PE-5712: Add allow listing by gateway operators for approving their delegated stakers

arielmelendez:
Adds the following:
1. The ability to specify the `Allow-Delegated-Staking` value during `Join-Network` as `allowlist`, which enables delegated staking for the members of a gateway's allow list ONLY.
2. The ability to specify allow-listed delegates during `Join-Network` via a comma-separated list of AO-valid delegator addresses in the `Allowed-Delegates` tag.
3. An `Allow-Delegates` handler that adds more allow-listed delegates via a comma-separated list of AO-valid delegator addresses in the `Allowed-Delegates` tag.
4. A `Disallow-Delegates` handler that removes allow-listed delegates via a comma-separated list of AO-valid delegator addresses in the `Disallowed-Delegates` tag.

Updates the following to account for allowlisting:
1. `gar.delegateStake`
2. `gar.increaseExistingStake`
3. `gar.decreaseDelegateStake`
4. `gar.assertValidGatewayParameters` to account for allowlisting parameters
5. `gar.pruneGateways`
6. `gar.cancelGatewayWithdrawal`
7. `gar.instantDelegateWithdrawal`

Implementation details:
- During `Join-Network` and `Update-Gateway-Settings`, if `Allow-Delegated-Staking` is set to `allowlist`, then a field is added to the gateway object called `allowedDelegatesLookup` - a table whose keys are delegate addresses for 0(1) lookup. The allowDelegatedStaking variable is still stored as true/false, but the passed in setting basically kicks the gateway into or out of allowlisting mode.
- When a delegate initially stakes, the allow list lookup is checked for permission, and if allowed, then the delegate is given a balance as normal and its name is removed from the allow list to save on memory. As such...
- A delegate is considered allowed if either: they are in the allow list lookup table OR the have an existing delegate balance above 0. A delegate with a vault and 0 balance and NOT in the allow list is considered disallowed and cannot cancel their withdrawal (i.e. attempt to reinstate their stake), increase their stake, or stake again.
- When a delegate withdraws stake, its address is only returned to the allow list at delegate pruning time (i.e. when all vaults are empty and delegated stake is 0).

TODO:
1. Update `updateGatewaySettings` (Handle toggling of allowDelegatedStaking) ✅ 
2. Integration tests ✅ 
3. Remove/replace all allow listed delegates somehow ✅ 
4. Decide whether disabling delegated staking purges the allow list ✅  (It does)
5. Decide whether gateway exit on slash purges the allow list ✅ (it does not)
6. Decide whether gateway exit by owner purges the allow list ✅  (it does not)


spec/utils_spec.lua
```diff
@@ -290,4 +290,94 @@ describe("utils", function()
...
+		it("should handle custom delimiter without trimming unexpected characters", function()
+			local input = "one two three"
+			local result = utils.splitAndTrimString(input, " ")
+			assert.are.same({ "one", "two", "three" }, result)
+		end)
+
+		it("should handle consecutive delimiters as separate items", function()
+			local input = "apple,,banana, ,cherry,"
+			local result = utils.splitAndTrimString(input)
+			assert.are.same({ "apple", "banana", "cherry" }, result)
```
fedellen, line 334: just imagining a dev just blasting commas all over a list to cause errors and then it just works instead 😂

src/arns.lua
```diff
@@ -433,7 +433,7 @@ function arns.upgradeRecord(from, name, currentTimestamp)
 	local demandFactor = demand.getDemandFactor()
 	local upgradeCost = arns.calculatePermabuyFee(baseFee, demandFactor)
 
-	if not utils.walletHasSufficientBalance(from, upgradeCost) then
+	if not balances.walletHasSufficientBalance(from, upgradeCost) then
```
dtfiedler, line 436: 👍

src/main.lua
```diff
@@ -2898,4 +2905,106 @@ addEventingHandler("auctionBid", utils.hasMatchingTag("Action", ActionMap.Auctio
...
+	local shouldContinue2, result = eventingPcall(msg.ioEvent, function(error)
+		ao.send({
+			Target = msg.From,
+			Tags = { Action = "Invalid-" .. ActionMap.AllowDelegates .. "-Notice", Error = "Bad-Input" },
+			Data = tostring(error),
+		})
+	end, gar.allowDelegates, newAllowedDelegates, msg.From)
+	if not shouldContinue2 then
+		return
+	end
```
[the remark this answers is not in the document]
dtfiedler: in lua - pcalls return a tuple. the first is if the called function succeeds without error (true/false). the second is the result of the function, or the error thrown by the function. you'll commonly see it as `resultOrError` or `maybeResult`.

tests/gar.test.mjs
```diff
@@ -775,329 +510,533 @@ describe('GatewayRegistry', async () => {
...
+          Object.keys(updatedGateway.delegates),
+          expectedDelegates,
+        );
+      }
+    }
+
+    it('should allow joining of the network with an allow list', async () => {
+      await allowlistJoinTest({
+        tags: [
+          { name: 'Allow-Delegated-Staking', value: 'allowlist' },
```
dtfiedler, line 540: just a reminder for ourselves - we'll want to update the SDK to support this value (and the new lookup table)

tests/gar.test.mjs
```diff
@@ -775,329 +510,533 @@ describe('GatewayRegistry', async () => {
...
+        ],
+        expectedAllowDelegatedStaking: false,
+      });
+    });
+
+    it('should join the network and start an empty allow list if Allow-Delegated-Staking is set to "allowlist"', async () => {
+      await allowlistJoinTest({
+        tags: [{ name: 'Allow-Delegated-Staking', value: 'allowlist' }],
+        expectedAllowDelegatedStaking: true,
+        expectedAllowedDelegatesLookup: [], // TODO: Why is an empty Lua table JSON serialized to an array?
```
dtfiedler, line 576: it's part of json.lua - without any objects, there's no way to clarify if you want a json object, or table back.

arielmelendez: Yep, figured this was the case!

spec/gar_spec.lua
```diff
@@ -516,6 +532,12 @@ describe("gar", function()
 					},
 				},
 			}
+			local expectedSettings = utils.deepCopy(testSettings)
+			expectedSettings.allowedDelegatesLookup = {
+				["test-allowlisted-delegator-address-number-1"] = true,
+				["test-allowlisted-delegator-address-number-2"] = true,
+				[stubRandomAddress] = true,
```
dtfiedler, line 539: why is this in the list?

dtfiedler (approved):
🚀

dtfiedler:
some patch coverage we can address in a separate PR - this one has gotten rather big 😄

dtfiedler (approved):

spec/gar_spec.lua
arielmelendez: Permission to delegate is indicated by either:
- presence in the allow list
- holding a delegated stake balance
When your stake is exited from a gateway, your permission is preserved by re-entering your in the allow list. In this case, since the delegate wasn't explicitly removed from the allow list, they were moved back into the allow list when their delegated stake was exited.

dtfiedler: got it, makes sense.

dtfiedler:
> Decide whether gateway exit on slash purges the allow list ⏱️
> Decide whether gateway exit by owner purges the allow list ⏱️

Hmm, thinking out loud. Right now, when a gateway exists (either slash or leaving) the following happens:
- operator stake is put in to vaults
- all existing delegated stake is vaulted
- gateway is set to "leaving"
- totalDelegatedStake and totalOperatorStake are set to 0

Things that do not happen:
- there is no resetting gateway stats (we leave them, as proof for why they were booted)
- changing any gateway settings (which this would be)

So given that, I don't think we need to worry about clearing out the allow list when leaving. It can be useful for 1. debugging and 2. visibility - operators could poach delegators that were assigned to a gateway now marked as leaving

src/main.lua
```diff
@@ -1711,10 +1716,7 @@ addEventingHandler(
 				or not disableDelegatedStaking -- NOT clear directive to DISABLE
 					and gateway.settings.allowDelegatedStaking, -- otherwise unspecified, so use previous setting
 
-			allowedDelegates = not shouldClearAllowlist
-					and (enableLimitedDelegatedStaking or gateway.settings.allowedDelegatesLookup)
-					and msg.Tags["Allowed-Delegates"]
-					and utils.splitAndTrimString(msg.Tags["Allowed-Delegates"]) -- replace the lookup list - TODO: REMOVE EXISTING DELEGATES
+			allowedDelegates = needNewAllowlist and utils.splitAndTrimString(msg.Tags["Allowed-Delegates"], ",") -- replace the lookup list - TODO: REMOVE EXISTING DELEGATES
```
dtfiedler, line 1719: is this still a TODO?

dtfiedler (approved):

arielmelendez:
> > Decide whether gateway exit on slash purges the allow list ⏱️
> > Decide whether gateway exit by owner purges the allow list ⏱️
> 
> Hmm, thinking out loud. Right now, when a gateway exists (either slash or leaving) the following happens:
> 
> * operator stake is vaulted
> * all delegated stake is vaulted
> * gateway is set to "leaving"
> * totalDelegatedStake and totalOperatorStake are set to 0
> 
> Things that do not happen:
> 
> * resetting gateway stats (we leave them as proof for why they were booted)
> * modifying/ressetting gateway settings (which this would be)
> 
> So, given that, I don't think it's necessary to clear out the allowed list when leaving. One example where it could be useful to keep - a gateway with an allow list is marked as leaving; operators are looking for delegators. They can review "leaving" gateways" and pursue those delegators for their stake - e.g add them to their allow list.
> 
> Can you think of any pro to clearing the list? Saving memory maybe?

I think there's no need to clear the allow list in those scenarios since the gateway is essentially unrecoverable and all staking operations are effectively cancelled. I've marked the open questions as closed there.

src/main.lua
arielmelendez: Nope! fba4dd4

src/gar.lua
```diff
@@ -937,48 +1002,135 @@ function gar.instantDelegateWithdrawal(from, gatewayAddress, vaultId, currentTim
...
+		gateway.settings.allowedDelegatesLookup[delegateAddress] = true
+	end
+end
+
+function gar.delegationAllowlistedOnGateway(gateway)
+	return gateway.settings.allowedDelegatesLookup ~= nil
+end
+
+function gar.delegateAllowedToStake(delegateAddress, gateway)
+	if not gar.delegationAllowlistedOnGateway(gateway) then
```
kunstmusik, line 1128: Noticed this function does not check gateway.settings.allowDelegatedStaking. I see in the code that we check for this outside of this function in places but seems safer to me to add a check here too (if there isn't going to be a later gas cost associated). Otherwise might be good to add a comment that gateway.settings.allowDelegatedStaking should be checked prior to using this function.

PR #161, opened 2024-11-02
fix(withdrawals): standardize cancel withdrawls and instant withdrawals, update tests

dtfiedler:
Similar to `Instant-Withdrawal`, `Cancel-Withdrawal` can use the same handler for delegates and operators. This avoids duplicate logic, and makes it easier to test edge cases. 

Note: this PR includes some large changes to integration tests. The intent is to create re-usable functions for doing things like `joinNetwork`, `delegateStake`, etc. for setting up state in tests without duplicattion. It should make the tests easier to read and extend for new edge cases.

Related PR: https://github.com/ar-io/ar-io-sdk/pull/246


tests/gar.test.mjs
```diff
@@ -49,41 +51,366 @@ describe('GatewayRegistry', async () => {
...
+    // assert no error tag
+    const errorTag = delegateResult.Messages?.[0]?.Tags?.find(
+      (tag) => tag.name === 'Error',
+    );
+    assert.strictEqual(errorTag, undefined);
+
+    return {
+      result: delegateResult,
+      memory: delegateResult.Memory,
+    };
```
dtfiedler, line 98: @arielmelendez - these functions wrap happy path handler behavior to set up the desired state easily within a test. See the updated tests for how they can be used to test various edge cases. The existing integration tests primarily focusing on validating happy path behavior of handlers, and inspecting various pieces of the resulting state (balances, gar, etc.). 

I'll look towards continuing to make integration test buildout easier across all test suites - this is just a starting point.

tests/gar.test.mjs
```diff
@@ -174,930 +537,371 @@ describe('GatewayRegistry', async () => {
...
+        vaultOwner: delegatorAddress,
+        gatewayAddress: STUB_ADDRESS,
+        vaultId: decreaseStakeMsgId,
+        memory: decreaseStakeMemory,
+        timestamp: decreaseStakeTimestamp,
+      });
+      // get the gateway record
+      const gatewayAfter = await getGateway({
+        address: STUB_ADDRESS,
+        memory: cancelWithdrawalMemory,
```
dtfiedler, line 801: @arielmelendez an example of setting up state to test a specific handler/edge case using the functions above

.vscode/settings.json
```diff
@@ -20,6 +20,10 @@
     "**/dist": true,
     "**/coverage": true
   },
+  "Lua.codeLens.enable": true,
```
arielmelendez, line 23: Sweet thanks for these

src/arns.lua
```diff
@@ -296,27 +305,24 @@ function arns.assertValidBuyRecord(name, years, purchaseType, processId)
 	end
 
 	-- assert processId is valid pattern
-	assert(type(processId) == "string", "ProcessId is required and must be a string.")
-	assert(utils.isValidBase64Url(processId), "ProcessId pattern is invalid.")
+	assert(type(processId) == "string", "Process id is required and must be a string.")
+	assert(utils.isValidAOAddress(processId), "Process id must be a valid base64url.")
```
arielmelendez, line 309: `a valid base64url` is somewhat ambiguous. Consider: `must be a valid AO signer address.`

src/arns.lua
```diff
@@ -296,27 +305,24 @@ function arns.assertValidBuyRecord(name, years, purchaseType, processId)
...
 end
 
+--- Asserts that a record is valid for extending the lease
+--- @param record table|nil The record to check
+--- @param currentTimestamp number|nil The current timestamp
```
arielmelendez, line 314: Why express that any of these can be nil when nil will cause an error to be thrown?

src/arns.lua
```diff
@@ -296,27 +305,24 @@ function arns.assertValidBuyRecord(name, years, purchaseType, processId)
...
-		error("Name is expired")
-	end
+	assert(record.type ~= "permabuy", "Name is permanently owned and cannot be extended")
+	assert(not arns.recordExpired(record, currentTimestamp), "Name is expired")
 
 	local maxAllowedYears = arns.getMaxAllowedYearsExtensionForRecord(record, currentTimestamp)
-	if years > maxAllowedYears then
-		error("Cannot extend lease beyond 5 years")
-	end
+	assert(years <= maxAllowedYears, "Cannot extend lease beyond 5 years")
```
arielmelendez, line 325: This assertion pattern is pretty clean.

src/arns.lua
```diff
@@ -353,6 +359,9 @@ function arns.getRegistrationFees()
 	return fees
 end
 
+--- Gets the token cost for an intended action
+--- @param intendedAction table The intended action
```
arielmelendez, line 363: Might be worth somehow specifying what the expected table keys are. I know the tooling likely won't pick it up, but it might still help the reader.

src/gar.lua
```diff
@@ -339,7 +349,7 @@ function gar.delegateStake(from, target, qty, currentTimestamp)
 		error("Cannot delegate to your own gateway, use increaseOperatorStake instead.")
 	end
 
-	if balances.getBalance(from) < qty then
+	if not balances.walletHasSufficientBalance(from, qty) then
```
arielmelendez, line 352: Could even save a few lines around the code with a `balances.assertSufficientBalance(from, qty)`

src/gar.lua
```diff
@@ -166,7 +173,10 @@ local function processInstantWithdrawal(qty, elapsedTime, totalWithdrawalTime, f
 	penaltyRate = math.max(
 		constants.MIN_EXPEDITED_WITHDRAWAL_PENALTY_RATE,
 		math.min(constants.MAX_EXPEDITED_WITHDRAWAL_PENALTY_RATE, penaltyRate)
-	) -- Ensure penalty is within bounds
+	)
+
+	-- round to three decimal places
```
arielmelendez, line 178: What's the thought process/spec here? Also might make sense to write a function and unit test for this.

src/gar.lua
```diff
@@ -156,6 +156,13 @@ function gar.increaseOperatorStake(from, qty)
 end
 
 -- Utility function to calculate withdrawal details and handle balance adjustments
+---@param qty number # The amount of stake to withdraw
```
arielmelendez, line 159: Nit: some params specify that the amounts are in mIO, others don't. Perhaps we need something in the readme that specifies that all currency amounts used throughout the codebase are in mIO unless otherwise specified (and perhaps even that final qualifier need not apply since it's all mIO all the time)...

src/gar.lua
```diff
@@ -156,6 +156,13 @@ function gar.increaseOperatorStake(from, qty)
...
+---@param elapsedTime number # The amount of time that has elapsed since the withdrawal started
```
arielmelendez, line 160: Rename this field to include the time unit (i.e. milliseconds)

arielmelendez: Same for `totalWithdrawalTime`.

src/gar.lua
```diff
@@ -495,7 +505,7 @@ function gar.decreaseDelegateStake(gatewayAddress, delegator, qty, currentTimest
 	-- update the gateway
 	GatewayRegistry[gatewayAddress] = gateway
 	return {
-		gateway = gar.getGateway(gatewayAddress),
+		gateway = gateway,
```
arielmelendez, line 508: Good change. Might want to search the codebase for other opportunities to save cycles like this.

src/gar.lua
dtfiedler: the penalty rate does not need precision greater than 3 digits as we round down the resulting mIO charged when a penalty occurs. avoids having to cast to string and potential loss of precision in tests.

src/gar.lua
dtfiedler: we should never be concerned with IO in the contract, but agree these parameter names (and generally around the code base) can be more accurate. as we continue to update docs, we can standardize.

tweaked these here - [8f13aa5](https://github.com/ar-io/ar-io-network-process/pull/161/commits/8f13aa515d4798eeb8fe2920e20814ba6df7b6d9)

src/gar.lua
dtfiedler: [8f13aa5](https://github.com/ar-io/ar-io-network-process/pull/161/commits/8f13aa515d4798eeb8fe2920e20814ba6df7b6d9)

src/gar.lua
dtfiedler: updated other uses here - [8f13aa5](https://github.com/ar-io/ar-io-network-process/pull/161/commits/8f13aa515d4798eeb8fe2920e20814ba6df7b6d9)

there are still other uses, but they are when we are operating on the resulting gateway to increment stats, change weights, etc. and the deep clone prevents unintentionally modifying the reference gateway before all modifications are made.

src/gar.lua
dtfiedler: moved to a util here - [8f13aa5](https://github.com/ar-io/ar-io-network-process/pull/161/commits/8f13aa515d4798eeb8fe2920e20814ba6df7b6d9)

added test here -

src/arns.lua
dtfiedler: tweaked here - [8550415](https://github.com/ar-io/ar-io-network-process/pull/161/commits/85504157f932d851172d0d028ec06f66be7f244e)

src/gar.lua
```diff
@@ -919,24 +929,41 @@ function gar.cancelGatewayWithdrawal(from, gatewayAddress, vaultId)
...
 	if gateway == nil then
 		error("Gateway not found")
 	end
 
-	local delegate = gateway.delegates[from]
-	if delegate == nil then
-		error("Delegate not found")
+	local isGatewayWithdrawal = from == gatewayAddress
+
+	if isGatewayWithdrawal and gateway.status == "leaving" then
```
arielmelendez, line 946: Can use assert

spec/utils_spec.lua
```diff
@@ -385,4 +385,31 @@ describe("utils", function()
...
+			local result = utils.roundToPrecision(-1.23456789, 3)
+			assert.are.equal(-1.235, result)
+		end)
+
+		it("should handle zero", function()
+			local result = utils.roundToPrecision(0, 3)
+			assert.are.equal(0, result)
+		end)
+
+		it("should handle large numbers", function()
```
arielmelendez, line 410: What's this specific case actually testing?

src/gar.lua
```diff
@@ -949,61 +977,23 @@ function gar.instantDelegateWithdrawal(from, gatewayAddress, vaultId, currentTim
...
-	-- Remove the delegate if no stake is left
-	if delegate.delegatedStake == 0 and next(delegate.vaults) == nil then
-		gar.pruneDelegateFromGateway(from, gateway)
+	if isGatewayWithdrawal then
+		gateway.vaults[vaultId] = nil
+	else
+		if delegate == nil then
+			error("Delegate not found")
+		end
+		delegate.vaults[vaultId] = nil
```
arielmelendez, line 986: This needs to use `gar.pruneDelegateFromGateway(from, gateway)` to jive with allowlist

src/gar.lua
```diff
@@ -949,61 +977,23 @@ function gar.instantDelegateWithdrawal(from, gatewayAddress, vaultId, currentTim
...
+		-- Remove the delegate if no stake is left
+		if delegate.delegatedStake == 0 and next(delegate.vaults) == nil then
+			gateway.delegates[from] = nil
+		end
 	end
 
 	-- Update the gateway
 	GatewayRegistry[gatewayAddress] = gateway
 	return {
-		delegate = gar.getGateway(gatewayAddress).delegates[from],
```
arielmelendez, line 962: nice

arielmelendez (changes requested):
Please fix delegate pruning as directed.

src/gar.lua
arielmelendez: 9e315d0

PR #163, opened 2024-11-05
docs: add steps for dev on-boarding

fedellen:
ran into a few hiccups during dev env setup. macOS brew install puts the include folder at `.../lua@5.3/include/lua5.3` rather than `.../lua@5.3/include`. In addition there are a few more installations needed for running the commands in the readme: `rust`, `luacov`, `aos`


dtfiedler (approved):
long overdue - ty!

PR #164, opened 2024-11-05
feat: init arns discount elgibility and fee caclulation PE-7060

fedellen:


src/gar.lua
```diff
@@ -981,4 +982,21 @@ function gar.instantOperatorWithdrawal(from, vaultId, currentTimestamp)
...
+	if gateway == nil then
+		return false
+	end
+
+	if gateway.weights == nil then
+		return false
+	end
+
+	local tenureWeight = gateway.weights.tenureWeight
+	local gatewayPerformanceRatio = gateway.weights.gatewayRewardRatioWeight
```
dtfiedler, line 996: nit: this shouldn't happen, but lua throws up when comparing numbers to `nil` so I would add `or 0` to both of these

dtfiedler (commented):
look good! 

look into adding an integration test to `gar.test.mjs` - you can use @arielmelendez [current branch](https://github.com/ar-io/ar-io-network-process/pull/157) for how to set up state with a gateway and then call handlers like `buy-record` and `get-token-cost` to validate the discount is applied for operators if they achieve the thresholds defined in the process. happy to pair on it speed up iterations.

src/arns.lua
```diff
@@ -502,8 +532,12 @@ function arns.upgradeRecord(from, name, currentTimestamp)
...
 
 	local baseFee = demand.baseFeeForNameLength(#name)
-	local demandFactor = demand.getDemandFactor()
-	local upgradeCost = arns.calculatePermabuyFee(baseFee, demandFactor)
+	local upgradeCost = arns.getTokenCost({
+		currentTimestamp = currentTimestamp,
+		intent = "Upgrade-Name",
+		name = name,
+		from = from,
+	})
```
dtfiedler, line 540: nice! been meaning to use this more internally - hopefully it's helpful

src/gar.lua
```diff
@@ -1109,6 +1110,23 @@ function gar.allowDelegates(delegates, gatewayAddress)
...
 
+function gar.isEligibleForArNSDiscount(from)
+	local gateway = gar.getGateway(from)
+	if gateway == nil then
+		return false
+	end
+
+	if gateway.weights == nil then
+		return false
+	end
```
dtfiedler, line 1121: nit, can combine to the same if
```suggestion
	if gateway == nil or gateay.weights == nil then
		return false
	end
```

src/main.lua
```diff
@@ -1013,6 +1013,7 @@ addEventingHandler(ActionMap.TokenCost, utils.hasMatchingTag("Action", ActionMap
 		quantity = tonumber(msg.Tags.Quantity),
 		purchaseType = msg.Tags["Purchase-Type"] or "lease",
 		currentTimestamp = tonumber(msg.Timestamp) or tonumber(msg.Tags.Timestamp),
+		from = msg.From,
```
dtfiedler, line 1016: 👏

dtfiedler:
In addition to the current changes, we can modify the return type of `getTokenCost` to include `discounts`. I believe @arielmelendez is updating it to account for `fundingSources` in this PR. https://github.com/ar-io/ar-io-network-process/pull/170

fedellen:
putting in draft. plan to merge this feature into buy_from_stake from the integrated branch. will close when that happens

PR #170, opened 2024-11-08
PE-7059: Buy records with various funding sources

arielmelendez:


src/main.lua
```diff
@@ -993,40 +1023,64 @@ addEventingHandler(ActionMap.TokenCost, utils.hasMatchingTag("Action", ActionMap
...
+		return
 	end
+
+	ao.send({
+		Target = msg.From,
+		Tags = { Action = "Token-Cost-Notice", ["Token-Cost"] = tostring(tokenCost) },
+		Data = fundFrom and json.encode({
+				tokenCost = tokenCost,
+				fundingPlan = fundingPlan,
+			})
```
dtfiedler, line 1080: mentioned in https://github.com/ar-io/ar-io-network-process/pull/172 - let's also include `discounts` as the result from `getTokenCost`

src/gar.lua
```diff
@@ -1191,4 +1196,259 @@ function gar.delegateAllowedToStake(delegateAddress, gateway)
...
+	return fundingPlan
+end
+
+function planBalanceDrawdown(fundingPlan, sourcesPreference)
+	local availableBalance = balances.getBalance(fundingPlan.address)
+	if sourcesPreference == "balance" or sourcesPreference == "any" then
+		fundingPlan.balance = math.min(availableBalance, fundingPlan.shortfall)
+		fundingPlan.shortfall = fundingPlan.shortfall - fundingPlan.balance
+	end
+end
```
dtfiedler, line 1240: note: bc these are not exported you likely cannot unit test

src/gar.lua
```diff
@@ -1191,4 +1196,259 @@ function gar.delegateAllowedToStake(delegateAddress, gateway)
...
+			delegatedStake = stakeToDraw,
+			vaults = {}, -- set up vault spend tracking now while we're passing through
+		}
+		fundingPlan.shortfall = fundingPlan.shortfall - stakeToDraw
+		nextGateway.delegate.delegatedStake = nextGateway.delegate.delegatedStake - stakeToDraw
+		-- maintain consistency
+		nextGateway.delegate.excessStake = excessStake - stakeToDraw
+		nextGateway.totalDelegatedStake = nextGateway.totalDelegatedStake - stakeToDraw
+		gatewayIndex, nextGateway = next(gatewaysInfo, gatewayIndex)
+	end
```
dtfiedler, line 1303: for readability's sake you could also use for loop to avoid abstract use of `next` and potential infinite while loop

src/gar.lua
```diff
@@ -1191,4 +1196,259 @@ function gar.delegateAllowedToStake(delegateAddress, gateway)
...
+		local stakeToDraw = math.min(excessStake, fundingPlan.shortfall)
+		fundingPlan["stakes"][nextGateway.delegate.gatewayAddress] = {
+			delegatedStake = stakeToDraw,
+			vaults = {}, -- set up vault spend tracking now while we're passing through
+		}
+		fundingPlan.shortfall = fundingPlan.shortfall - stakeToDraw
+		nextGateway.delegate.delegatedStake = nextGateway.delegate.delegatedStake - stakeToDraw
+		-- maintain consistency
+		nextGateway.delegate.excessStake = excessStake - stakeToDraw
+		nextGateway.totalDelegatedStake = nextGateway.totalDelegatedStake - stakeToDraw
```
dtfiedler, line 1301: can you explain why this? are you updating the reference to the current gateway to avoid double counting the stake available from it? more detailed comment might be helpful

src/gar.lua
```diff
@@ -1191,4 +1196,259 @@ function gar.delegateAllowedToStake(delegateAddress, gateway)
...
+			end
+			return acc
+		end, {}),
+		{
+			{
+				order = "asc",
+				field = "endTimestamp",
+			},
+		}
+	)
```
dtfiedler, line 1327: the nested sort/reduce are hard to follow. may be helpful to breakup, at the sacrifice of memory footprint

src/gar.lua
arielmelendez: Correct - the are tested in integration with the top level function.

src/gar.lua
arielmelendez: Dang for some reason I convinced myself that you can't break out of a for loop in lua 🤔 But you can. I will make that adjustment.

src/gar.lua
dtfiedler: ahhh i see - makes sense

src/gar.lua
arielmelendez: 4d392ad

src/gar.lua
arielmelendez: Expanded comment:
98f6b38

src/gar.lua
arielmelendez: b7f254d

PR #174, opened 2024-11-11
Arns-Discount-buy-from-stake-integrated

fedellen:


tests/gar.test.mjs
```diff
@@ -1331,10 +1331,10 @@ describe('GatewayRegistry', async () => {
...
         if (!cursor) break;
       }
-      assert.deepEqual(
-        fetchedGateways.map((g) => g.gatewayAddress),
-        [STUB_ADDRESS, secondGatewayAddress],
-      );
+
+      fetchedGateways.forEach((g) => {
+        assert([STUB_ADDRESS, secondGatewayAddress].includes(g.gatewayAddress));
+      });
```
fedellen, line 1337: can remove, fixed in above brnach

src/gar.lua
```diff
@@ -1111,6 +1112,19 @@ function gar.allowDelegates(delegates, gatewayAddress)
 	}
 end
 
+function gar.isEligibleForArNSDiscount(from)
+	local gateway = gar.getGateway(from)
```
arielmelendez, line 1116: Could introduce `getGatewayUnsafe` and not suffer the compute and memory costs of cloning all delegate stakes and vaults since this is a read-only operation.

src/constants.lua
```diff
@@ -21,6 +21,9 @@ constants.UNDERNAME_PERMABUY_FEE_PERCENTAGE = 0.005
 constants.oneYearMs = 31536000 * 1000
 constants.gracePeriodMs = 14 * 24 * 60 * 60 * 1000 -- 2 weeks
 constants.maxLeaseLengthYears = 5
+constants.ARNS_DISCOUNT_PERCENTAGE = 0.2
+constants.ARNS_DISCOUNT_TENURE_WEIGHT_ELIGIBILITY_FACTOR = 1
+constants.ARNS_DISCOUNT_GATEWAY_PERFORMANCE_RATIO_ELIGIBILITY_FACTOR = 0.85
```
arielmelendez, line 26: Consider `THRESHOLD` over `FACTOR`

src/main.lua
```diff
@@ -971,6 +971,7 @@ addEventingHandler(
 		local fundFrom = msg.Tags["Fund-From"]
 		local checkAssertions = function()
 			assert(type(msg.Tags.Name) == "string", "Invalid name")
+			assert(msg.Tags.Quantity, "Must provide a quantity")
```
arielmelendez, line 974: Nit: This is sort of already considered on the next line. Consider simply adding `(msg.Tags.Quantity and ...` to the line below if nil-ness is the problem.

src/arns.lua
```diff
@@ -453,7 +486,8 @@ end
 ---   - name string The name of the record
 ---   - intent string The intended action type (Buy-Record/Extend-Lease/Increase-Undername-Limit/Upgrade-Name)
 ---   - currentTimestamp number The current timestamp
---- @return number The token cost in mIO of the intended action
+---   - from string|nil The target address of the intended action
+--- @return TokenCostResult The token cost in mIO of the intended action
 function arns.getTokenCost(intendedAction)
```
arielmelendez, line 491: I kinda wonder if we should rename this now that it returns more than just a cost. It's creating a little bit of readability friction in the code.

tests/arns.test.mjs
```diff
@@ -133,6 +133,8 @@ describe('ArNS', async () => {
     };
   };
 
+  const basePermabuyPrice = 2500000000;
```
arielmelendez, line 136: Thanks for this. Wouldn't hate some underscores to separate the 3-digit clusters.

tests/arns.test.mjs
```diff
@@ -1509,4 +1512,237 @@ describe('ArNS', async () => {
   });
 
   // TODO: add several error scenarios
+
+  describe('ArNS Discount Eligible Gateways', () => {
+    const getBalances = async ({ memory }) => {
```
arielmelendez, line 1517: Might make for a nice helper function.

tests/arns.test.mjs
```diff
@@ -1509,4 +1512,237 @@ describe('ArNS', async () => {
...
+      // TODO: Join failing gateway -- Submit an observation with another joined gateway that marks the gateway as failing
+      // const failingGateway = 'failing-unique-gateway-'.padEnd(43, '0');
+      // const { memory: join2Memory } = await joinNetwork(
+      //   {
+      //     address: failingGateway,
+      //   },
+      //   firstTick.Memory,
+      // );
+
+      // TODO: We should be able to create and distribute an epoch within one tick. But for now, this errors -- so we tick twice
```
arielmelendez, line 1551: What's the nature of the error?

arielmelendez (approved):
Approved with a few suggestions.

src/arns.lua
```diff
@@ -494,11 +528,31 @@ function arns.getTokenCost(intendedAction)
 		arns.assertValidUpgradeName(record, currentTimestamp)
 		tokenCost = arns.calculatePermabuyFee(baseFee, demand.getDemandFactor())
 	end
+
+	local discounts = {}
+
+	-- if the address is eligible for the ArNS discount, apply the discount
+	if gar.isEligibleForArNSDiscount(intendedAction.from) then
+		local discountedCost = math.floor(tokenCost * constants.ARNS_DISCOUNT_PERCENTAGE)
```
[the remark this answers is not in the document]
dtfiedler: agree - would call this `discountQty` or `discountTotal`

src/main.lua
```diff
@@ -1108,15 +1112,22 @@ addEventingHandler(ActionMap.TokenCost, utils.hasMatchingTag("Action", ActionMap
 		return
 	end
 
+	local data
+	if fundFrom ~= nil or discounts ~= nil then
+		data = {
+			tokenCost = tokenCost,
+			fundingPlan = fundFrom and fundingPlan or nil,
+			discounts = discounts,
+		}
```
dtfiedler, line 1121: nit - I suggest we introduce a new handler that supports the new object returned by this function. this will give us backwards compatibility as we integrate into arns.app and SDK. 

propose the handler name is `getCostDetailsForAction` or something similar

tests/arns.test.mjs
fedellen: I may have solved this error in another PR where `prevEpoch.distributions` was nil. Will try this sgain

src/main.lua
fedellen: for this refactor, would we leave discounts out of this return object and only put it into the new getCostDetailsForAction return? 

fwiw we can start to gracefully expect object or raw tokenCost on the getTokenCost SDK methods so I'm not super worried about backwards compatibility. but the new handler would make sense so the return is always expected

tests/arns.test.mjs
fedellen: e034170

tests/arns.test.mjs
fedellen: bb1deb5

src/main.lua
fedellen: Without this line we throw an aos type error below

src/constants.lua
fedellen: 17ebb53

src/arns.lua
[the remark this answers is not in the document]
fedellen: b78f6f3

src/gar.lua
fedellen: 8320fc7

src/main.lua
dtfiedler: i would keep this to only return the raw token cost, the new handler can return funding plan and disconts

tests/arns.test.mjs
fedellen: still doesnt work. specifically the first tick won't cause any distributions, which results in the gateway not receiving tenure

```   'Ticking epoch: 0\n' +
      'Not updating demand factor\n' +
      'Not distributing rewards for last epoch.\n' +
      'Distributions have not occured for the previous epoch. A new epoch will not be created until those are complete: 0\n' +
      'Ticking epoch: 1\n' +
      'Not updating demand factor\n' +
      'Not distributing rewards for last epoch.\n' +
      'Distributions have not occured for the previous epoch. A new epoch will not be created until those are complete: 0\n' +
      '{"Delegated-Supply":0,"New-Demand-Factors":[1,1],"Cron":false,"Message-Id":"mmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm","_e":1,"Leaving-Gateways-Count":0,"Staked-Supply":0,"Action":"Tick","Locked-Supply":0,"Current-Epoch-Index":0,"Target-Current-Epoch-Index":1,"From-Formatted":"1111111111111111111111111111111111111111111","Circulating-Supply":0,"Epoch-Index":1,"Joined-Gateways-Count":0,"Total-Token-Supply":50000000000000,"From":"1111111111111111111111111111111111111111111","Withdraw-Supply":0,"Protocol-Balance":50000000000000,"Cast":false,"Last-Ticked-Epoch-Index":-1,"Ticked-Epoch-Indexes":[0],"Timestamp":1719988800000}',```

File: src/gar.lua [hunk 11]
```diff
@@ -1111,6 +1116,19 @@ function gar.allowDelegates(delegates, gatewayAddress)
 	}
 end
 
+function gar.isEligibleForArNSDiscount(from)
+	local gateway = gar.getGatewayUnsafe(from)
+	if gateway == nil or gateway.weights == nil then
```
dtfiedler, line 1121: i would check that gateway is active as well. leaving gateways should not get discounts.

```suggestion
	if gateway == nil or gateway.weights == nil or gar.isGatewayLeaving(gateway) then
```

src/main.lua
dtfiedler: you can modify to account for both

```suggestion
			assert( msg.Tags.Quantity and 
				tonumber(msg.Tags.Quantity) > 0
					and tonumber(msg.Tags.Quantity) < 9990
					and utils.isInteger(msg.Tags.Quantity),
				"Invalid quantity. Must be an integer value greater than 0 and less than 9990"
			)```

Comment (fedellen) [reply src/main.lua]: 8c31c61

File: tests/arns.test.mjs (hunk 8 above)
Comment (dtfiedler) [reply tests/arns.test.mjs:1540]: does it work if you use the distribution timestamp to create AND distribute the first epoch?

```
      const futureTickResult = await handle(
        {
          Tags: [{ name: 'Action', value: 'Tick' }],
          Timestamp: afterDistributionTimestamp,
        },
        firstTick.Memory,
      );
```

File: tests/tick.test.mjs [hunk 12]
```diff
@@ -320,10 +320,8 @@ describe('Tick', async () => {
     const futureTimestamp = vaultData.endTimestamp + 1;
     const futureTick = await handle(
       {
-        Tags: [
-          { name: 'Action', value: 'Tick' },
-          { name: 'Timestamp', value: futureTimestamp.toString() },
-        ],
+        Tags: [{ name: 'Action', value: 'Tick' }],
+        Timestamp: futureTimestamp,
```
dtfiedler, line 324: ty

src/gar.lua
fedellen: ee8d8f3

tests/arns.test.mjs
```diff
@@ -1509,4 +1513,225 @@ describe('ArNS', async () => {
...
+    const firstEpochTimestamp = 1719900000000;
+    const afterDistributionTimestamp =
+      firstEpochTimestamp + 1000 * 60 * 60 * 24 + 1000 * 60 * 40;
+
+    let arnsDiscountMemory;
+    before(async () => {
+      // create gateways
+      const { memory: join1Memory } = await joinNetwork({
+        address: joinedGateway,
+        quantity: 200_000_000_000,
```
dtfiedler, line 1528: make sure this timestamp equal to or less than the firstEpochTimestamp

src/main.lua
fedellen: 77de4b1

tests/arns.test.mjs
fedellen: yes, the above comment has the log result with that setup. the gateway does not receive tenure in that case

src/main.lua
```diff
@@ -1108,15 +1112,22 @@ addEventingHandler(ActionMap.TokenCost, utils.hasMatchingTag("Action", ActionMap
 		return
 	end
 
+	local data
+	if fundFrom ~= nil or discounts ~= nil then
```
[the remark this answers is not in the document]
fedellen: doesn't exeist here anymore -- this has been refactored in get-cost-details-for-action

tests/arns.test.mjs
fedellen: 3c6a2dc

src/arns.lua
arielmelendez: This commit just seems to update docs.

tests/arns.test.mjs
fedellen: had to add a followup to get this to pass: 32cd72e

src/arns.lua
```diff
@@ -653,6 +704,12 @@ function arns.submitAuctionBid(name, bidAmount, bidder, timestamp, processId, ty
 
 	local finalBidAmount = math.min(requiredOrBidAmount, requiredBid)
```
arielmelendez, line 705: Nit: this doesn't need to be math.min - it's always just the requiredBid.

tests/arns.test.mjs
```diff
@@ -1509,4 +1513,226 @@ describe('ArNS', async () => {
...
+        });
+        const priceOfOneUnderName = 500_000;
+        const expectedCost = priceOfOneUnderName * 20 * 0.8;
+        assert.equal(
+          balancesAfter[joinedGateway],
+          balancesBefore[joinedGateway] - expectedCost,
+        );
+      });
+    });
+  });
```
arielmelendez, line 1737: TODO: Add a test for submitAuctionBid case. Just call anything like "getBalances" after the grace period is over (which is weeks PAST the endTimestamp of the record) and the prepend handler will cause an auction to get created. Then test pricing out the auction bid.

arielmelendez:
Update the auction price handler to factor in discounts somehow:
```
addEventingHandler("auctionPrices", utils.hasMatchingTag("Action", ActionMap.AuctionPrices), function(msg)
```
e.g. another field on the return data like the discount rate that the client can use to apply/show discounts.

src/arns.lua
```diff
@@ -653,6 +704,12 @@ function arns.submitAuctionBid(name, bidAmount, bidder, timestamp, processId, ty
...
 
+	-- check if bidder is eligible for ArNS discount
+	if gar.isEligibleForArNSDiscount(bidder) then
```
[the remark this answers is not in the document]
fedellen: d206577

tests/arns.test.mjs
fedellen: 8c35a96eac5e1df8dd966fe499b1d04af0e9d89e

src/arns.lua
fedellen: within https://github.com/ar-io/ar-io-network-process/commit/d206577db7372e566ad3aa4ba5493769cff561b5

src/main.lua
```diff
@@ -1091,35 +1091,111 @@ addEventingHandler(ActionMap.TokenCost, utils.hasMatchingTag("Action", ActionMap
...
-			or json.encode(tokenCost),
+		Data = json.encode(tokenCost),
 	})
 end)
 
+addEventingHandler(
+	ActionMap.GetCostDetailsForAction,
+	utils.hasMatchingTag("Action", ActionMap.GetCostDetailsForAction),
+	function(msg)
+		local fundFrom = msg.Tags["Fund-From"]
```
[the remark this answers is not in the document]
fedellen: did this partially with joining the assertions together: bbc1a3b

could continue further in the file, but more broadly asserting in prepend handler might work better on afoolow up

dtfiedler (approved):

PR #181, opened 2024-11-12
test(tick): init coverage on base fee coming down after going up PE-6998

fedellen:


tests/tick.test.mjs
```diff
@@ -659,4 +659,102 @@ describe('Tick', async () => {
...
+          Tags: [
+            { name: 'Action', value: 'Buy-Record' },
+            { name: 'Name', value: 'test-name-' + i },
+            { name: 'Purchase-Type', value: 'permabuy' },
+          ],
+        },
+        buyRecordMemory,
+      );
+      buyRecordMemory = buyRecordResult.Memory;
+    }
```
dtfiedler, line 708: nit: you don't have to buy records to validate the base fees drop after consecutive periods at the minimum

I recommend splitting this up into two tweets
1. checks fees during an epoch, buy a few records, and check base fees again (they should not change) and validate the demand factor has increased correctly
2. check base fees at first epoch, tick the consecutive number of epochs required to reduce base registration fees without any name purchases, validate the base fees dropped correctly compared to the original

fedellen: 86060a2

tests/tick.test.mjs
```diff
@@ -659,4 +664,121 @@ describe('Tick', async () => {
...
+        },
+        tickMemory,
+      );
+
+      tickMemory = Memory;
+    }
+
+    const baseFeeAfterConsecutiveTicksWithNoPurchases =
+      await getBaseRegistrationFee(tickMemory);
+    assert.equal(baseFeeAfterConsecutiveTicksWithNoPurchases, 573_402_975);
```
dtfiedler, line 782: add check that the demand factor was reset to 1 after adjusting fees

fedellen: demand factor after 4 ticks is `0.95567162500000002456`

fedellen: for 20 ticks, it is `0.750392318118884738`

fedellen: we need 49 ticks from genesis to get back to 1 demand factor. the fee remains at 300_000_000 however

image

91d1ba0

tests/helpers.mjs
```diff
@@ -158,3 +158,27 @@ export const setUpStake = async ({
     result: stakeResult,
   };
 };
+
+export const getBaseRegistrationFee = async ({ memory, timestamp }) => {
```
dtfiedler, line 162: nit: provide the name and then compute the length to return the name - becomes `getBaseRegistrtionFeeForName`

tests/tick.test.mjs
```diff
@@ -657,4 +662,156 @@ describe('Tick', async () => {
...
+      if ([46, 47, 48].includes(i)) {
+        const demandFactor = await getDemandFactor({
+          memory: tickMemory,
+          timestamp: epochTimestamp,
+        });
+        assert.equal(demandFactor, 0.5);
+      }
+    }
+
+    const afterTimestamp = genesisEpochStart + (epochDurationMs + 1) * 50;
```
arielmelendez, line 803: Should this be:
```
genesisEpochStart + epochDurationMs * 50 + 1
```
?

arielmelendez (dismissed):

tests/tick.test.mjs
fedellen: its implied with * 50 rather than 49. but setting the timestamps here don't seem to change the outcome --  the memory seems to be the only important variable for getDemandFactor and getBaseFees

fedellen: on the 49th tick since genesis we reset and return back to demandFactor 1

tests/helpers.mjs
fedellen: 0a253b6

PR #185, opened 2024-11-12
chore: add git hash to evolve.mjs PE-6589

fedellen:
accidently pushed to develop, verified this during that:

image

### DIFF
--- a/tools/evolve.mjs
+++ b/tools/evolve.mjs
@@ -1,6 +1,16 @@
 import { AOProcess, createAoSigner, ArweaveSigner } from '@ar.io/sdk/node';
 import * as constants from './constants.mjs';
 import { connect } from '@permaweb/aoconnect';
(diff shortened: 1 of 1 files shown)


tools/evolve.mjs
```diff
@@ -1,6 +1,16 @@
...
 import { connect } from '@permaweb/aoconnect';
+import { execSync } from 'child_process';
+
+// Retrieve the Git hash directly in the script
+let gitHash;
+try {
+  gitHash = execSync('git rev-parse --short HEAD').toString().trim();
+} catch (error) {
+  console.error('Error retrieving Git hash:', error);
+  gitHash = 'unknown';
```
dtfiedler, line 12: nit: don't include if not defined vs. unkown

dtfiedler (approved):

PR #186, opened 2024-11-13
PE-7093: delegations handler

arielmelendez:
The handler returns a heterogenous, flattened list of active and vaulted delegations. Sorting can be done somewhat effectively by any fields that are common to both types of item. Here's an example list of items:
```
      [
          {
            type: 'vault',
            gatewayAddress: '2222222222222222222222222222222222222222222',
            startTimestamp: 21600002,
            delegationId:
              '2222222222222222222222222222222222222222222_21600002',
            balance: 400000001,
            vaultId: 'decrease-stake-message-id',
            endTimestamp: 2613600002,
          },
          {
            type: 'stake',
            gatewayAddress: '2222222222222222222222222222222222222222222',
            delegationId:
              '2222222222222222222222222222222222222222222_21600000',
            balance: 599999999,
            startTimestamp: 21600000,
          },
          {
            type: 'stake',
            gatewayAddress: 'second-gateway-aaaaaaaaaaaaaaaaaaaaaaaaaaaa',
            delegationId:
              'second-gateway-aaaaaaaaaaaaaaaaaaaaaaaaaaaa_21600001',
            balance: 600000000,
            startTimestamp: 21600001,
          },
        ]
```
The `delegationId` field was added to eliminate ambiguity of cursor values that exist for other fields. For example, timestamps might not be an effective field because a pruning operation could kick 2 gateways at which a delegate is staked, resulting in 2 withdraw vaults with the same startTimestamp and message Id.

### DIFF
--- a/spec/gar_spec.lua
+++ b/spec/gar_spec.lua
@@ -2676,4 +2676,155 @@ describe("gar", function()
 			}, _G.GatewayRegistry["gateway-2"])
 		end)
 	end)
+
+	describe("getPaginatedDelegations", function()
+		local gateway1 = utils.deepCopy(testGateway)
+		local gateway2 = utils.deepCopy(testGateway)
+		gateway1.delegates = {
+			["test-user"] = {
+				delegatedStake = 1,
+				startTimestamp = 2,
+				vaults = {
+					["vault_id_1"] = {
+						balance = 1000,
+						startTimestamp = 4,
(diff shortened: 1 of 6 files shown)


dtfiedler (commented):
let's add the API and type to a branch of the ar-io-sdk and review the UX for this handler

PR #187, opened 2024-11-13
fix(index): avoid indexing by number on gateway slashings

dtfiedler:


src/main.lua
```diff
@@ -304,19 +304,19 @@ end, function(msg)
...
+		-- 	for name, record in pairs(resultOrError.newGracePeriodRecords) do
+		-- 		table.insert(newGracePeriodRecordNames, name)
+		-- 		if record.endTimestamp > LastGracePeriodEntryEndTimestamp then
+		-- 			LastGracePeriodEntryEndTimestamp = record.endTimestamp
+		-- 		end
+		-- 	end
+		-- 	msg.ioEvent:addField("New-Grace-Period-Records", newGracePeriodRecordNames)
+		-- 	msg.ioEvent:addField("New-Grace-Period-Records-Count", newGracePeriodRecordsCount)
+		-- 	msg.ioEvent:addField("Last-Grace-Period-Entry-End-Timestamp", LastGracePeriodEntryEndTimestamp)
+		-- end
```
dtfiedler, line 319: this seemed to cause 
```
Pruning state at timestamp: 1731498271246
ERROR: Invalid field value type: table. Supported types are string, number, boolean, or nil.
Ticking epoch: 127
Not updating demand factor
Rewards already distributed for epoch: 126
Epoch already exists for index: 127
```
not sure if it's tied to this event data, or something else

dtfiedler:
manually applied to testnet - will close and push commits to fix via `develop`

PR #189, opened 2024-11-13
fix(discounts): recs for discount price handler and tests

dtfiedler:


src/main.lua
```diff
@@ -1102,96 +1105,68 @@ addEventingHandler(ActionMap.TokenCost, utils.hasMatchingTag("Action", ActionMap
 			from = msg.From,
 		}
 	)
-	if not shouldContinue2 then
+	if not shouldContinue2 or not tokenCostResult then
```
fedellen, line 1108: also had this there for awhile, but was thinking it'd be nice if we could make the type def of `eventingPCall` handle this the same way `pcall` does. in TS its like `returns [false, undefined] | [true, T: tokenCostResult]`.

src/epochs.lua
```diff
@@ -386,9 +386,9 @@ function epochs.createEpoch(timestamp, blockHeight, hashchain)
 
 	local prevEpochIndex = epochIndex - 1
 	local prevEpoch = epochs.getEpoch(prevEpochIndex)
-	-- if the previous epoch has not distributed rewards, we cannot create a new epoch
+	-- if the previous epoch is not the genesis epoch and we have not distributed rewards, we cannot create a new epoch
 	if
-		prevEpochIndex >= 0
+		prevEpochIndex > 0 -- only validate distributions occured if previous epoch is not the genesis epoch
```
fedellen, line 391: 🎉

fedellen (approved):

PR #192, opened 2024-11-14
feat(redelegate stake): init gar implementaion PE-7058

fedellen:
TODO:

- ~~redelegate from vaults (do we put into a vault, or as a delegation)~~
- ~~add pruning of redelegates when seven epochs reached~~
- ~~finish unit tests~~
- ~~init integration tests~~
  - extend integration tests
- ~~getReDelegationFee handler~~
- ~~Handle event API -- caclucate delegated stake~~
- better PR description
- more maybe


src/main.lua
```diff
@@ -1559,6 +1560,86 @@ addEventingHandler(ActionMap.DelegateStake, utils.hasMatchingTag("Action", Actio
...
+	if gateway ~= nil then
+		local newStake = gateway.delegates[from].delegatedStake
+		msg.ioEvent:addField("Previous-Stake", newStake - quantity)
+		msg.ioEvent:addField("New-Stake", newStake)
+		msg.ioEvent:addField("Gateway-Total-Delegated-Stake", gateway.totalDelegatedStake)
+		delegateResult = gateway.delegates[from]
+	end
+
+	LastKnownDelegatedSupply = LastKnownDelegatedSupply + quantity
+	addSupplyData(msg.ioEvent)
```
fedellen, line 1634: todo: check this, co-pilot generated stuff without context

fedellen: will exercise this handler when setting up integration tests

src/main.lua
```diff
@@ -2380,7 +2382,11 @@ addEventingHandler(ActionMap.Gateways, Handlers.utils.hasMatchingTag("Action", A
 end)
 
 addEventingHandler(ActionMap.Gateway, Handlers.utils.hasMatchingTag("Action", ActionMap.Gateway), function(msg)
+	print("Gateway action")
+	print(msg.Tags.Address)
+	print(msg.From)
 	local gateway = gar.getGateway(msg.Tags.Address or msg.From)
+
```
fedellen, line 2389:
```suggestion
	local gateway = gar.getGateway(msg.Tags.Address or msg.From)
```

src/gar.lua
```diff
@@ -1577,4 +1577,271 @@ function gar.getPaginatedDelegations(address, cursor, limit, sortBy, sortOrder)
...
+		gar.delegateAllowedToStake(delegateAddress, targetGateway),
+		"This Gateway does not allow this delegate to stake."
+	)
+
+	local previousRedelegations = gar.getRedelegation(delegateAddress)
+
+	local redelegationFeeRate = gar.getRedelegationFee(delegateAddress).redelegationFeeRate
+
+	local redelegationFee = math.ceil(stakeToTakeFromSource * (redelegationFeeRate / 100))
+	local stakeToDelegate = stakeToTakeFromSource - redelegationFee
```
arielmelendez, line 1675: So the redelegator chooses an amount of stake they are willing to participate in the move, and they basically get that amount minus any fees redelegated?

The other approach would be to move the amount they want to move and assess fees on TOP of that. Just want to make sure we're aligned with the intention from the white paper.

src/gar.lua
```diff
@@ -1577,4 +1577,271 @@ function gar.getPaginatedDelegations(address, cursor, limit, sortBy, sortOrder)
...
+			assert(
+				existingDelegate.delegatedStake >= stakeToTakeFromSource,
+				"Quantity must be less than or equal to the delegated stake amount."
+			)
+
+			-- Check if the delegate will have enough stake left after re-delegating
+			local existingStake = existingDelegate.delegatedStake
+			local requiredMinimumStake = sourceGateway.settings.minDelegatedStake
+			local maxAllowedToWithdraw = existingStake - requiredMinimumStake
+			assert(
```
arielmelendez, line 1752: It's interesting that we have this barrier here since the user could always commence a withdraw of their stake and then begin a redelegation. We grapple with this a bit with funding plans where, since we acknowledge this potential user workaround, we err on the side of a good UX for the user and simply "break a 20" on their minimum stake if necessary. We'll use it all if we have to, or we'll leave behind the excess in a withdraw vault. I'm wondering why we wouldn't just do the same here (we can check with Jon).

src/gar.lua
```diff
@@ -1577,4 +1577,271 @@ function gar.getPaginatedDelegations(address, cursor, limit, sortBy, sortOrder)
...
+				stakeToTakeFromSource <= maxAllowedToWithdraw or stakeToTakeFromSource == existingStake,
+				"Remaining delegated stake must be greater than the minimum delegated stake. Adjust the amount or re-delegate all stake."
+			)
+
+			-- If the delegate has enough stake to redelegate, move the stake. If its all the stake, remove the delegate
+			if existingDelegate.delegatedStake == stakeToTakeFromSource then
```
arielmelendez, line 1758: This is only appropriate if the delegate has no withdraw vaults. We'll have to fix this.

src/gar.lua
```diff
@@ -1577,4 +1577,271 @@ function gar.getPaginatedDelegations(address, cursor, limit, sortBy, sortOrder)
...
+			end
+			sourceGateway.totalDelegatedStake = sourceGateway.totalDelegatedStake - stakeToTakeFromSource
+		end
+	end
+
+	local existingTargetDelegate = targetGateway.delegates[delegateAddress]
+	local minimumStakeForGatewayAndDelegate
+	if existingTargetDelegate and existingTargetDelegate.delegatedStake ~= 0 then
+		-- It already has a stake that is not zero
+		minimumStakeForGatewayAndDelegate = 1 -- Delegate must provide at least one additional IO
```
arielmelendez, line 1772: Is this supposed to be 1 IO or 1 mIO?

src/gar.lua
```diff
@@ -1577,4 +1577,271 @@ function gar.getPaginatedDelegations(address, cursor, limit, sortBy, sortOrder)
...
+
+	-- Move redelegation fee to protocol balance
+	balances.increaseBalance(ao.id, redelegationFee)
+
+	Redelegations[delegateAddress] = {
+		timestamp = currentTimestamp,
+		redelegations = (previousRedelegations and previousRedelegations.redelegations or 0) + 1,
+	}
+
+	-- prune user from allow list, if necessary, to save memory
```
arielmelendez, line 1809: Not quite how this works. User goes back on the allowlist on the source gateway IF they've been pruned, otherwise their active balance or vaults "holds their spot" on the gateway. On the destination gateway, we take them out of the allowlist lookup if we create a new delegation for them.

src/gar.lua
arielmelendez: Confirmed with Jon that this was the intended flow 👍

src/gar.lua
arielmelendez: I believe it should say mIO. I've changed the comment this was copied from in the delegate stake flow and will update here as well.

src/gar.lua
arielmelendez: Standardized this with a few utility functions in the gar. Should be all good now.

src/gar.lua
arielmelendez: Fixed

PR #198, opened 2024-11-18
chore(main): remove handlers that are unbounded or old

dtfiedler:
All large pieces of state should be returned as paginated arrays.

Need to validate e2e tests with SDK continue to work with current return types on these updated handlers.

### DIFF
--- a/src/main.lua
+++ b/src/main.lua
@@ -1947,18 +1947,9 @@ addEventingHandler(
 	utils.hasMatchingTag("Action", "Paginated-Records") or utils.hasMatchingTag("Action", ActionMap.Records),
 	function(msg)
 		local page = utils.parsePaginationTags(msg)
-		local status, result =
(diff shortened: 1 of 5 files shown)


fedellen (approved):

src/main.lua
```diff
@@ -1961,7 +1908,8 @@ addEventingHandler(ActionMap.Distributions, utils.hasMatchingTag("Action", Actio
 end)
 
 addEventingHandler(ActionMap.ReservedNames, utils.hasMatchingTag("Action", ActionMap.ReservedNames), function(msg)
```
arielmelendez, line 1910: Nit: the other paginated handlers have paginated in the name...

fedellen (dismissed):

src/main.lua
dtfiedler: [c28921d](https://github.com/ar-io/ar-io-network-process/pull/198/commits/c28921d519a53e9cd568164c1215a19c31f085ce)

arielmelendez (approved):

PR #205, opened 2024-11-19
fix(error): throw error on failures, return event data in error message

dtfiedler:


src/ao_event.lua
```diff
@@ -94,7 +94,12 @@ local function AOEvent(initialData)
 	function event:printEvent()
 		-- The _e: 1 flag signifies that this is an event. Ensure it is set.
 		self.data["_e"] = 1
```
arielmelendez, line 96: Nit: this is redundant since it takes place in toJSON now

dtfiedler: [23a1558](https://github.com/ar-io/ar-io-network-process/pull/205/commits/23a15584c9a2bfb3f1342448eb6dfad3bc167b23)

arielmelendez (approved):

PR #206, opened 2024-11-19
fix(tick): move tickEpoch to separate function

dtfiedler:
We will add unit tests to make sure this happens correctly

### DIFF
--- a/src/main.lua
+++ b/src/main.lua
@@ -33,6 +33,7 @@ local demand = require("demand")
 local epochs = require("epochs")
 local vaults = require("vaults")
 local prune = require("prune")
+local tick = require("tick")
 local primaryNames = require("primary_names")
 
 local ActionMap = {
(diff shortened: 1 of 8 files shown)


src/main.lua
```diff
@@ -1574,36 +1576,11 @@ end)
...
-		local newEpoch = epochs.createEpoch(timestamp, tonumber(blockHeight), hashchain)
-		return {
-			maybeEpoch = newEpoch,
-			maybeDemandFactor = demandFactor,
-			pruneGatewaysResult = pruneGatewaysResult,
-		}
-	end
 
 	local msgId = msg.Id
+	local blockHeight = tonumber(msg["Block-Height"])
```
arielmelendez, line 1582: Should be a tag?

arielmelendez: And also already tonumbered?

src/main.lua
```diff
@@ -1574,36 +1576,11 @@ end)
...
+	local hashchain = msg["Hash-Chain"]
```
arielmelendez, line 1583: tag?

src/main.lua
dtfiedler: This is a top level attribute on all messages, similar to timestamp

fedellen (dismissed):

PR #207, opened 2024-11-19
chore: devnet to testnet

dtfiedler:
Includes:
- paginated reserved arns name handler
- primary name support
- standardized input formatting for addresses, quantities, etc.
- stake redelegation support
- removed `vaults` and `delegates` from gateway handlers, added separate handlers with pagination
- removed unused code
- minimum vault size of 100 IO
- updated error handling for `prune` and `distribute`
   - these functions will throw out memory on failures to avoid potentially corrupted state
- aos 2.0.1 support - `.update` be applied manually before merging

Related SDK PR: https://github.com/ar-io/ar-io-sdk/pull/261


.vscode/settings.json
```diff
@@ -36,5 +36,6 @@
     "editor.formatOnSave": true,
     "editor.formatOnPaste": true,
     "editor.formatOnSaveMode": "file"
-  }
+  },
+  "cSpell.words": ["redelegate"]
```
fedellen, line 40: can remove as we're not maintaining a list here. I think I added via the IDE it out of habit

dtfiedler: fine leaving, if you see more feel free to add.

fedellen (dismissed):

PR #214, opened 2024-11-20
feat(vaults): skip vault pruning work when possible PE-7147

arielmelendez:


src/vaults.lua
```diff
@@ -6,6 +6,9 @@ local balances = require("balances")
 local utils = require("utils")
 local constants = require("constants")
 
+--- @type Timestamp|nil
+NextPruneTimestamp = NextPruneTimestamp or 0
```
dtfiedler, line 10: will this be the global timestamp for all pruning activity? if so, consider moving to main.lua

dtfiedler: also may want to consider if we pass this down to prune functions rather than referencing the global timestamp everywhere.

arielmelendez: a056f58

src/vaults.lua
```diff
@@ -38,6 +41,9 @@ function vaults.createVault(from, qty, lockLengthMs, currentTimestamp, vaultId)
 		startTimestamp = currentTimestamp,
 		endTimestamp = currentTimestamp + lockLengthMs,
 	})
+	-- A nil NextPruneTimestamp means we're not expecting anything to prune, so set it if necessary
+	-- Otherwise, this new endTimestamp might be earlier than the next known for pruning. If so, set it.
+	NextPruneTimestamp = math.min(NextPruneTimestamp or newVault.endTimestamp, newVault.endTimestamp)
```
dtfiedler: this leaves the responsibility of the function to update global state. is there any pro to having the callers of `createVault` do it instead ?

src/vaults.lua
```diff
@@ -183,17 +193,31 @@ end
...
 				prunedVaults[id] = nestedVault
+			else
+				--- find the next prune timestamp
+				minNextEndTimestamp =
+					math.min(minNextEndTimestamp or nestedVault.endTimestamp, nestedVault.endTimestamp)
 			end
 		end
 	end
+
+	NextPruneTimestamp = minNextEndTimestamp
```
dtfiedler: similar comment, should `pruneVaults` return the minimum timestamp back to `prune.lua` and it set the NextMinimum? curious your thoughts

src/vaults.lua
arielmelendez: > also may want to consider if we pass this down to prune functions rather than referencing the global timestamp everywhere.

The global is the lightest lift at the moment. Otherwise I'll have to collect all the outputs of every usage of the functions that might mutate the next timestamp value and apply that to a different global that lives somewhere else. Need to globals to survive evals.

src/vaults.lua
arielmelendez: Makes the APIs harder to use correctly IMO if you shift the responsibility outside.

src/vaults.lua
dtfiedler: with the intent of using a global variable per pruneable collection of state (vaults, vs. records, vs. gateways) this comment may not apply

dtfiedler (approved):

PR #215, opened 2024-11-21
fix(aos): update handler to be compliant with 2.0.1 token spec

dtfiedler:
Also removes unused pnp supply data as we immediately withdraw from balance on primary name requests instead of holding balance.

### DIFF
--- a/src/main.lua
+++ b/src/main.lua
@@ -42,7 +42,7 @@ local CRITICAL = true
 local ActionMap = {
 	-- reads
 	Info = "Info",
-	TotalTokenSupply = "Total-Token-Supply",
+	TotalSupply = "Total-Supply",
 	State = "State",
 	Transfer = "Transfer",
 	Balance = "Balance",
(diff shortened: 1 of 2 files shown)


src/main.lua
```diff
@@ -1509,93 +1504,90 @@ addEventingHandler(
...
-	for _, gateway in pairs(gar.getGatewaysUnsafe()) do
-		totalSupply = totalSupply + gateway.operatorStake + gateway.totalDelegatedStake
-		stakedSupply = stakedSupply + gateway.operatorStake
-		delegatedSupply = delegatedSupply + gateway.totalDelegatedStake
-		for _, delegate in pairs(gateway.delegates) do
-			-- tally delegates' vaults
-			for _, vault in pairs(delegate.vaults) do
+addEventingHandler(
+	"totalTokenSupply",
+	utils.hasMatchingTag("Action", ActionMap.TotalSupply) or utils.hasMatchingTag("Action", "Total-Token-Supply"), -- TODO: remove this once we migrate all downstream apps to the new tag
```
dtfiedler, line 1509: we'll update the SDK to use the new tag and push that out in v2.5.0

src/main.lua
```diff
@@ -1952,61 +1955,52 @@ end)
...
-	"paginatedVaults",
-	utils.hasMatchingTag("Action", "Paginated-Vaults") or utils.hasMatchingTag("Action", ActionMap.Vaults),
-	function(msg)
-		local page = utils.parsePaginationTags(msg)
-		local pageVaults = vaults.getPaginatedVaults(page.cursor, page.limit, page.sortOrder, page.sortBy)
-		ao.send({ Target = msg.From, Action = "Vaults-Notice", Data = json.encode(pageVaults) })
-	end
-)
+addEventingHandler("paginatedVaults", function(msg)
+	return msg.Action == "Paginated-Vaults" or msg.Action == ActionMap.Vaults
```
dtfiedler, line 1984: this was a bug before - the old handler was not working

PR #216, opened 2024-11-21
fix(io): allow unsafe addresses

dtfiedler:
We allow optionally providing `Allow-Unsafe-Addresses` flag to `transfer` and `vaultedTransfer`. This gives flexibility to future signature types but provides some contract enforcement of address types by default. Enforcement of other address-like things - like `processId`'s - are relaxed. 

This also removes the protection of names that look like wallet addresses, but I can be convinced to leave that in.

### DIFF
--- a/src/arns.lua
+++ b/src/arns.lua
@@ -483,7 +483,6 @@ function arns.assertValidBuyRecord(name, years, purchaseType, processId)
 	assert(type(name) == "string", "Name is required and must be a string.")
 	assert(#name >= 1 and #name <= 51, "Name pattern is invalid.")
 	assert(name:match("^%w") and name:match("%w$") and name:match("^[%w-]+$"), "Name pattern is invalid.")
-	assert(not utils.isValidAOAddress(name), "Name cannot be a wallet address.")
 
 	-- assert purchase type if present is lease or permabuy
 	assert(purchaseType == nil or purchaseType == "lease" or purchaseType == "permabuy", "Purchase-Type is invalid.")
@@ -499,7 +498,7 @@ function arns.assertValidBuyRecord(name, years, purchaseType, processId)
 
 	-- assert processId is valid pattern
 	assert(type(processId) == "string", "Process id is required and must be a string.")
-	assert(utils.isValidAOAddress(processId), "Process Id must be a valid AO signer address..")
+	assert(utils.isValidAddress(processId), "Process Id must be a valid address.")
 end
 
 --- Asserts that a record is valid for extending the lease
@@ -1005,7 +1004,7 @@ end
 function arns.assertValidReassignName(record, currentTimestamp, from, newProcessId)
 	assert(record, "Name is not registered")
 	assert(currentTimestamp, "Timestamp is required")
-	assert(utils.isValidAOAddress(newProcessId), "Invalid Process-Id")
+	assert(utils.isValidAddress(newProcessId), "Invalid Process-Id")
 	assert(record.processId == from, "Not authorized to reassign this name")
 
 	if record.endTimestamp then
--- a/src/arns.lua
+++ b/src/arns.lua
@@ -7,6 +7,9 @@ local arns = {}
 local Auction = require("auctions")
 local gar = require("gar")
 
+--- @type Timestamp|nil
+NextRecordsPruneTimestamp = NextRecordsPruneTimestamp or 0
+
 --- @class NameRegistry
 --- @field reserved table<string, ReservedName> The reserved names
 --- @field records table<string, Record> The records
@@ -131,6 +134,10 @@ function arns.addRecord(name, record)
 	-- remove reserved name if it exists in reserved
 	if arns.getReservedName(name) then
 		NameRegistry.reserved[name] = nil
+	end
+
+	if record.endTimestamp then
+		arns.scheduleNextRecordsPrune(record.endTimestamp)
 	end
 end
 
@@ -404,6 +411,8 @@ function arns.modifyRecordEndTimestamp(name, newEndTimestamp)
 	local maxEndTimestamp = record.startTimestamp + maxLeaseLength
 	assert(newEndTimestamp <= maxEndTimestamp, "Cannot extend lease beyond 5 years")
 	NameRegistry.records[name].endTimestamp = newEndTimestamp
+	-- Guard against the invariant case where record may not expire sooner
+	arns.scheduleNextRecordsPrune(newEndTimestamp)
 	return arns.getRecord(name)
 end
 
@@ -754,6 +763,7 @@ function arns.upgradeRecord(from, name, currentTimestamp, msgId, fundFrom)
 	demand.tallyNamePurchase(upgradeCost)
 
 	record.endTimestamp = nil
+	-- figuring out the next prune timestamp would require a full scan of all records anyway so don't reschedule
 	record.type = "permabuy"
 	record.purchasePrice = upgradeCost
 
@@ -968,6 +978,13 @@ function arns.pruneRecords(currentTimestamp, lastGracePeriodEntryEndTimestamp)
 	lastGracePeriodEntryEndTimestamp = lastGracePeriodEntryEndTimestamp or 0
 	local prunedRecords = {}
 	local newGracePeriodRecords = {}
+	if not NextRecordsPruneTimestamp or NextRecordsPruneTimestamp > currentTimestamp then
+		return prunedRecords, newGracePeriodRecords
+	end
+
+	--- @type Timestamp|nil
+	local minNextEndTimestamp
+
 	-- identify any records that are leases and that have expired, account for a one week grace period in seconds
 	for name, record in pairs(arns.getRecords()) do
 		if arns.recordExpired(record, currentTimestamp) then
@@ -978,7 +995,19 @@ function arns.pruneRecords(currentTimestamp, lastGracePeriodEntryEndTimestamp)
 			and record.endTimestamp > lastGracePeriodEntryEndTimestamp
 		then
 			newGracePeriodRecords[name] = record
+			-- Make sure we prune when the grace period is over
+			arns.scheduleNextRecordsPrune(record.endTimestamp + constants.gracePeriodMs)
+		elseif record.endTimestamp then
+			-- find the next prune timestamp
+			--- @diagnostic disable-next-line: param-type-mismatch
+			minNextEndTimestamp = math.min(minNextEndTimestamp or record.endTimestamp, record.endTimestamp)
 		end
+	end
+
+	-- Reset the next pruning timestamp now that pruning has completed
+	NextRecordsPruneTimestamp = nil
+	if minNextEndTimestamp then
+		arns.scheduleNextRecordsPrune(minNextEndTimestamp)
 	end
 	return prunedRecords, newGracePeriodRecords
 end
@@ -1043,6 +1072,11 @@ function arns.reassignName(name, from, currentTimestamp, newProcessId)
 	arns.assertValidReassignName(record, currentTimestamp, from, newProcessId)
 	local updatedRecord = arns.modifyProcessId(name, newProcessId)
 	return updatedRecord
+end
+
+--- @param timestamp Timestamp
+function arns.scheduleNextRecordsPrune(timestamp)
+	NextRecordsPruneTimestamp = math.min(NextRecordsPruneTimestamp or timestamp, timestamp)
 end
 
 return arns
--- a/src/arns.lua
+++ b/src/arns.lua
@@ -474,15 +474,21 @@ function arns.calculateYearsBetweenTimestamps(startTimestamp, endTimestamp)
 	return yearsRemainingFloat
 end
 
+--- Asserts that a name is a valid ARNS name
+--- @param name string The name to check
+function arns.assertValidArNSName(name)
+	assert(type(name) == "string", "Name is required and must be a string.")
+	assert(#name >= constants.MIN_NAME_LENGTH and #name <= constants.MAX_NAME_LENGTH, "Name pattern is invalid.")
+	assert(name:match(constants.ARNS_NAME_REGEX), "Name pattern is invalid.")
+end
+
 --- Asserts that a buy record is valid
 --- @param name string The name of the record
 --- @param years number|nil The number of years to check
 --- @param purchaseType string|nil The purchase type to check
 --- @param processId string|nil The processId of the record
 function arns.assertValidBuyRecord(name, years, purchaseType, processId)
-	assert(type(name) == "string", "Name is required and must be a string.")
-	assert(#name >= 1 and #name <= 51, "Name pattern is invalid.")
-	assert(name:match("^%w") and name:match("%w$") and name:match("^[%w-]+$"), "Name pattern is invalid.")
+	arns.assertValidArNSName(name)
 
 	-- assert purchase type if present is lease or permabuy
 	assert(purchaseType == nil or purchaseType == "lease" or purchaseType == "permabuy", "Purchase-Type is invalid.")
--- a/src/arns.lua
+++ b/src/arns.lua
@@ -418,10 +418,11 @@ end
 
 ---Calculates the lease fee for a given base fee, years, and demand factor
 --- @param baseFee number The base fee for the name
---- @param years number The number of years
+--- @param years number|nil The number of years
 --- @param demandFactor number The demand factor
 --- @return number leaseFee - the lease fee
 function arns.calculateLeaseFee(baseFee, years, demandFactor)
+	assert(years, "Years is required for lease")
 	local annualRegistrationFee = arns.calculateAnnualRenewalFee(baseFee, years)
 	local totalLeaseCost = baseFee + annualRegistrationFee
 	return math.floor(demandFactor * totalLeaseCost)
--- a/src/arns.lua
+++ b/src/arns.lua
@@ -477,9 +477,16 @@ end
 --- Asserts that a name is a valid ARNS name
 --- @param name string The name to check
 function arns.assertValidArNSName(name)
-	assert(type(name) == "string", "Name is required and must be a string.")
-	assert(#name >= constants.MIN_NAME_LENGTH and #name <= constants.MAX_NAME_LENGTH, "Name pattern is invalid.")
-	assert(name:match(constants.ARNS_NAME_REGEX), "Name pattern is invalid.")
+	assert(name and type(name) == "string", "Name is required and must be a string.")
+	assert(
+		#name >= constants.MIN_NAME_LENGTH and #name <= constants.MAX_NAME_LENGTH,
+		"Name length is invalid. Must be between "
+			.. constants.MIN_NAME_LENGTH
+			.. " and "
+			.. constants.MAX_NAME_LENGTH
+			.. " characters."
+	)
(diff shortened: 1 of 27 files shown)


src/arns.lua
```diff
@@ -474,16 +474,21 @@ function arns.calculateYearsBetweenTimestamps(startTimestamp, endTimestamp)
...
 --- Asserts that a buy record is valid
 --- @param name string The name of the record
 --- @param years number|nil The number of years to check
 --- @param purchaseType string|nil The purchase type to check
 --- @param processId string|nil The processId of the record
 function arns.assertValidBuyRecord(name, years, purchaseType, processId)
-	assert(type(name) == "string", "Name is required and must be a string.")
-	assert(#name >= 1 and #name <= 51, "Name pattern is invalid.")
-	assert(name:match("^%w") and name:match("%w$") and name:match("^[%w-]+$"), "Name pattern is invalid.")
-	assert(not utils.isValidAOAddress(name), "Name cannot be a wallet address.")
```
dtfiedler, line 486: we added this in, but bc we want to support more sig types in the future, it seems this should just be removed. fine to add back if you think otherwise @arielmelendez / @vilenarios

src/main.lua
```diff
@@ -1206,8 +1215,8 @@ end)
 addEventingHandler(ActionMap.CancelWithdrawal, utils.hasMatchingTag("Action", ActionMap.CancelWithdrawal), function(msg)
 	local gatewayAddress = msg.Tags.Target or msg.Tags.Address or msg.From
 	local vaultId = msg.Tags["Vault-Id"]
-	assert(utils.isValidAOAddress(gatewayAddress), "Invalid gateway address")
-	assert(utils.isValidAOAddress(vaultId), "Invalid vault id")
+	assert(utils.isValidAddress(gatewayAddress), "Invalid gateway address")
```
dtfiedler, line 1218: these are laxed bc to join the network `msg.From` will be used for the gateway address. if AO introduces support for new addresses, we don't want to worry about blocking delegations on those new addresses. the check for the `gateway` itself is sufficient.

src/main.lua
```diff
@@ -1427,10 +1436,10 @@ addEventingHandler(ActionMap.ReassignName, utils.hasMatchingTag("Action", Action
 	local initiator = msg.Tags.Initiator
 	local timestamp = msg.Timestamp
 	assert(name and #name > 0, "Name is required")
-	assert(utils.isValidAOAddress(newProcessId), "Process Id must be a valid AO signer address..")
+	assert(utils.isValidAddress(newProcessId), "Process Id must be a valid address.")
```
dtfiedler, line 1439: similarly - process ids can be flexible - maybe integration with EVM smart contracts will be available in the future 🤷

src/main.lua
```diff
@@ -540,6 +548,7 @@ addEventingHandler(ActionMap.Transfer, utils.hasMatchingTag("Action", ActionMap.
 			Action = "Debit-Notice",
 			Recipient = recipient,
 			Quantity = msg.Tags.Quantity,
+			["Allow-Unsafe-Addresses"] = tostring(allowUnsafeAddresses),
```
arielmelendez, line 551: Is this logically significant?

fedellen (dismissed):
I like the compromise here. Other iteration seemed a little dangerous for funds being lost due to typos or incomplete commands, although client-side could've still enforced address validation.

src/main.lua
dtfiedler: [e7e4e0a](https://github.com/ar-io/ar-io-network-process/pull/216/commits/e7e4e0a6e5d0305401b706cd5d3b8be13764628e)

src/main.lua
arielmelendez: Is it actually relaxed if you're not passing "allow unsafe" here?

src/main.lua
arielmelendez: Is it flexible if allow unsafe is not passed here?

src/main.lua
dtfiedler: 🤔

src/main.lua
dtfiedler: yeah you're right. frick me.

dtfiedler: i updated the call sig to not allow optionally, and set `true` where want to allow it to be relaxed.

[5601bc4](https://github.com/ar-io/ar-io-network-process/pull/216/commits/5601bc436f37046b22fd2ff71e366efc599fcb30)

lmk if you reason about it differently - my brain turned into spaghetti as we discussed the correct defaults for this

fedellen (dismissed):

src/arns.lua
```diff
@@ -474,16 +474,21 @@ function arns.calculateYearsBetweenTimestamps(startTimestamp, endTimestamp)
 	return yearsRemainingFloat
 end
 
+--- Asserts that a name is a valid ARNS name
+--- @param name string The name to check
+function arns.assertValidArNSName(name)
+	assert(type(name) == "string", "Name is required and must be a string.")
+	assert(#name >= constants.MIN_NAME_LENGTH and #name <= constants.MAX_NAME_LENGTH, "Name pattern is invalid.")
```
arielmelendez, line 481: Nit: could have a more descriptive error. Same for the regex pattern below. Could even have the regex string in the error message.

src/arns.lua
```diff
@@ -499,7 +504,7 @@ function arns.assertValidBuyRecord(name, years, purchaseType, processId)
 
 	-- assert processId is valid pattern
 	assert(type(processId) == "string", "Process id is required and must be a string.")
-	assert(utils.isValidAOAddress(processId), "Process Id must be a valid AO signer address..")
+	assert(utils.isValidAddress(processId, true), "Process Id must be a valid address.")
```
arielmelendez, line 507: Nit: could make a `utils.isValidArNSTarget` function to consolidate the logic here.

src/constants.lua
```diff
@@ -14,6 +14,14 @@ constants.minimumWithdrawalAmount = constants.mIOPerIO -- the minimum amount tha
...
 
 -- ARNS
+constants.MAX_NAME_LENGTH = 51
+constants.MIN_NAME_LENGTH = 1
+-- Regex pattern to validate ARNS names:
+-- - Starts with an alphanumeric character (%w)
+-- - Can contain alphanumeric characters and hyphens (%w-)
+-- - Ends with an alphanumeric character (%w)
+-- - Does not allow names to start or end with a hyphen
+constants.ARNS_NAME_REGEX = "^%w[%w-]*%w$"
```
arielmelendez, line 24: This regex does not allow 1 character names!

arielmelendez: I think it could be:
```
^%w[%w-]*%w?$
```

arielmelendez: Is there not a unit test ensuring that 1 character names are allowed?

dtfiedler: ha, looks like that is in fact missing. will add and reasses

src/arns.lua
dtfiedler: [f4db59a](https://github.com/ar-io/ar-io-network-process/pull/216/commits/f4db59a3098c6ab86484ec7beeacb9d08412cd92)

fedellen (dismissed):

fedellen (dismissed):

src/arns.lua
```diff
@@ -512,7 +524,7 @@ function arns.assertValidBuyRecord(name, years, purchaseType, processId)
...
```
arielmelendez, line 527: I think this might be one where we want to force users to specify allowing unsafe because if they get this wrong, they can brick the name. e.g. they wouldn't be able to use the ANT to reassign to another process ID.

src/arns.lua
```diff
@@ -1058,7 +1070,7 @@ end
 function arns.assertValidReassignName(record, currentTimestamp, from, newProcessId)
 	assert(record, "Name is not registered")
 	assert(currentTimestamp, "Timestamp is required")
-	assert(utils.isValidAOAddress(newProcessId), "Invalid Process-Id")
+	assert(utils.isValidAddress(newProcessId, true), "Invalid Process-Id") -- allow unsafe addresses for processId
```
arielmelendez, line 1073: Same comment here - we should force user to specify unsafe.

src/main.lua
```diff
@@ -572,15 +581,17 @@ addEventingHandler(ActionMap.Transfer, utils.hasMatchingTag("Action", ActionMap.
 			Action = "Debit-Notice",
 			Recipient = recipient,
 			Quantity = tostring(quantity),
-			Data = "You transferred " .. tostring(quantity) .. " to " .. recipient,
+			["Allow-Unsafe-Addresses"] = allowUnsafeAddresses,
```
arielmelendez, line 584: Does this need to be converted back to a string first?

src/main.lua
```diff
@@ -572,15 +581,17 @@ addEventingHandler(ActionMap.Transfer, utils.hasMatchingTag("Action", ActionMap.
...
+			Data = "You transferred " .. msg.Tags.Quantity .. " to " .. recipient,
 		}
 		-- Credit-Notice message template, that is sent to the Recipient of the transfer
 		local creditNotice = {
 			Target = recipient,
 			Action = "Credit-Notice",
 			Sender = msg.From,
 			Quantity = tostring(quantity),
-			Data = "You received " .. tostring(quantity) .. " from " .. msg.From,
+			["Allow-Unsafe-Addresses"] = allowUnsafeAddresses,
```
arielmelendez, line 593: Same comment here about converting back to a string first.

src/main.lua
```diff
@@ -678,7 +690,7 @@ addEventingHandler(ActionMap.VaultedTransfer, utils.hasMatchingTag("Action", Act
 		Target = msg.From,
 		Recipient = recipient,
 		Quantity = quantity,
-		Tags = { Action = "Debit-Notice", ["Vault-Id"] = msgId },
+		Tags = { Action = "Debit-Notice", ["Vault-Id"] = msgId, ["Allow-Unsafe-Addresses"] = allowUnsafeAddresses },
```
arielmelendez, line 693: Same issue here with converting back to string.

src/main.lua
```diff
@@ -756,15 +769,13 @@ addEventingHandler(ActionMap.BuyRecord, utils.hasMatchingTag("Action", ActionMap
...
 	assert(
 		type(purchaseType) == "string" and purchaseType == "lease" or purchaseType == "permabuy",
 		"Invalid purchase type"
 	)
 	assert(timestamp, "Timestamp is required for a tick interaction")
-	assert(type(name) == "string" and #name > 0 and #name <= 51 and not utils.isValidAOAddress(name), "Invalid name")
-	assert(type(processId) == "string", "Process id is required and must be a string.")
-	assert(utils.isValidAOAddress(processId), "Process Id must be a valid AO signer address..")
+	arns.assertValidArNSName(name)
+	assert(utils.isValidAddress(processId, true), "Process Id must be a valid address.")
```
arielmelendez, line 778: Mentioned elsewhere that this should for user to opt into unsafe since this could brick a name.

src/main.lua
```diff
@@ -1460,10 +1470,10 @@ addEventingHandler(ActionMap.ReassignName, utils.hasMatchingTag("Action", Action
 	local initiator = msg.Tags.Initiator
 	local timestamp = msg.Timestamp
 	assert(name and #name > 0, "Name is required")
-	assert(utils.isValidAOAddress(newProcessId), "Process Id must be a valid AO signer address..")
+	assert(utils.isValidAddress(newProcessId, true), "Process Id must be a valid address.")
```
arielmelendez, line 1473: Mentioned elsewhere, this should be OPT IN to unsafe to avoid bricking ANT.

src/main.lua
```diff
@@ -2019,9 +2029,9 @@ addEventingHandler("releaseName", utils.hasMatchingTag("Action", ActionMap.Relea
 	local initiator = msg.Tags.Initiator or msg.From
 	local timestamp = msg.Timestamp
 
-	assert(name and #name > 0, "Name is required")
-	assert(processId and utils.isValidAOAddress(processId), "Process-Id is required")
-	assert(initiator and utils.isValidAOAddress(initiator), "Initiator is required")
+	assert(name and #name > 0, "Name is required") --- this could be an undername, so we don't want to assertValidArNSName
```
arielmelendez, line 2032: Nit: don't need three hyphens

src/main.lua
```diff
@@ -2019,9 +2029,9 @@ addEventingHandler("releaseName", utils.hasMatchingTag("Action", ActionMap.Relea
...
+	assert(processId and utils.isValidAddress(processId, true), "Process-Id must be a valid address")
+	assert(initiator and utils.isValidAddress(initiator, true), "Initiator is required")
 	assert(record, "Record not found")
```
arielmelendez, line 2035: Perhaps we shouldn't be fetching the record before validating the name.

src/main.lua
```diff
@@ -2197,8 +2207,8 @@ addEventingHandler("auctionBid", utils.hasMatchingTag("Action", ActionMap.Auctio
 
 	-- assert name, bidder, processId are provided
 	assert(name and #name > 0, "Name is required")
-	assert(bidder and utils.isValidAOAddress(bidder), "Bidder is required")
-	assert(processId and utils.isValidAOAddress(processId), "Process-Id is required")
+	assert(bidder and utils.isValidAddress(bidder, true), "Bidder is required")
+	assert(processId and utils.isValidAddress(processId, true), "Process-Id must be a valid Arweave address")
```
arielmelendez, line 2211: Should be opt in to unsafe to avoid bricking names.

src/utils.lua
```diff
@@ -246,7 +246,7 @@ end
 
 --- Checks if an address is a valid Arweave address
 --- @param address string The address to check
---- @return boolean isValidArweaveAddress - whether the address is a valid Arweave address
+--- @return boolean isValidAddress - whether the address is a valid Arweave address
```
arielmelendez, line 249: No need for a variable name in the return doc. Just use a hashtag.

src/utils.lua
```diff
@@ -258,11 +258,27 @@ function utils.isValidEthAddress(address)
 	return type(address) == "string" and #address == 42 and string.match(address, "^0x[%x]+$") ~= nil
 end
 
+function utils.isValidUnsafeAddress(address)
+	if not address then
+		return false
+	end
+	local match = string.match(address, "^[%w_-]+$")
+	return match ~= nil and #address >= 1 and #address <= 128
```
arielmelendez, line 266: Nit: use constants

src/utils.lua
```diff
@@ -258,11 +258,27 @@ function utils.isValidEthAddress(address)
...
+end
+
 --- Checks if an address is a valid AO address
---- @param url string|nil The address to check
---- @return boolean isValidAOAddress - whether the address is a valid AO address
-function utils.isValidAOAddress(url)
-	return url and (utils.isValidArweaveAddress(url) or utils.isValidEthAddress(url)) or false
+--- @param address string|nil The address to check
+--- @param allowUnsafe boolean Whether to allow unsafe addresses, defaults to false
```
arielmelendez, line 271: type is `boolean|nil`

src/vaults.lua
```diff
@@ -52,8 +51,10 @@ end
 --- @param lockLengthMs number The lock length in milliseconds
 --- @param currentTimestamp number The current timestamp
 --- @param vaultId string The vault id
+--- @param allowUnsafeAddresses boolean Whether to allow unsafe addresses, since this results in funds eventually being sent to an invalid address
```
arielmelendez, line 54: Can relax the type to `boolean|nil`

arielmelendez (changes requested):
Needs a little work. Almost there though.

src/arns.lua
arielmelendez: f602db8

src/arns.lua
arielmelendez: e1f740d

src/main.lua
arielmelendez: e1f740d

src/utils.lua
arielmelendez: b2da18c

src/utils.lua
arielmelendez: c3992b1

src/utils.lua
arielmelendez: b2da18c

src/main.lua
arielmelendez: b2da18c

src/main.lua
arielmelendez: b2da18c

src/main.lua
arielmelendez: b2da18c

src/main.lua
arielmelendez: 616e10d

src/main.lua
arielmelendez: 616e10d

src/main.lua
arielmelendez: 616e10d

src/main.lua
```diff
@@ -689,6 +701,7 @@ addEventingHandler(ActionMap.VaultedTransfer, utils.hasMatchingTag("Action", Act
 		Tags = {
 			Action = ActionMap.CreateVault .. "-Notice",
 			["Vault-Id"] = msgId,
+			["Allow-Unsafe-Addresses"] = allowUnsafeAddresses,
```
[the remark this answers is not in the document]
arielmelendez: 616e10d

src/arns.lua
```diff
@@ -615,7 +631,7 @@ function arns.getTokenCost(intendedAction)
 	if intent == "Buy-Record" then
 		-- stub the process id as it is not required for this intent
 		local processId = "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
-		arns.assertValidBuyRecord(name, years, purchaseType, processId)
+		arns.assertValidBuyRecord(name, years, purchaseType, processId, false)
```
fedellen, line 634: this blocks getting a token cost for an unsafe name, we could use the allow-unsafe tag in this handler

fedellen: or might be better to have `true` in this read level API

src/main.lua
```diff
@@ -756,23 +773,32 @@ addEventingHandler(ActionMap.BuyRecord, utils.hasMatchingTag("Action", ActionMap
...
 	assert(
 		type(purchaseType) == "string" and purchaseType == "lease" or purchaseType == "permabuy",
 		"Invalid purchase type"
 	)
 	assert(timestamp, "Timestamp is required for a tick interaction")
-	assert(type(name) == "string" and #name > 0 and #name <= 51 and not utils.isValidAOAddress(name), "Invalid name")
-	assert(type(processId) == "string", "Process id is required and must be a string.")
-	assert(utils.isValidAOAddress(processId), "Process Id must be a valid AO signer address..")
+	arns.assertValidArNSName(name)
+	assert(utils.isValidAddress(processId, true), "Process Id must be a valid address.")
```
fedellen, line 783:
```suggestion
	arns.assertValidArNSName(name)
	assert(utils.isValidAddress(processId, allowUnsafeProcessId), "Process Id must be a valid address.")
```

fedellen: I guess not really, as long as it gets asserted again with the boolean passed down in `arns.lua`

fedellen (approved):

arielmelendez (approved):

PR #217, opened 2024-11-21
feat(gar): prune gateways only when necessary PE-7147

arielmelendez:


src/gar.lua
```diff
@@ -838,6 +847,11 @@ function gar.updateGatewayStats(address, gateway, stats)
 	assert(stats.failedConsecutiveEpochs, "failedConsecutiveEpochs is required")
 	assert(stats.passedConsecutiveEpochs, "passedConsecutiveEpochs is required")
 
+	if stats.failedConsecutiveEpochs >= gar.getSettings().operators.failedEpochCountMax then
+		-- Ensure pruning takes place at the next opportunity
+		NextGatewaysPruneTimestamp = 0
+	end
```
dtfiedler, line 853: this happens on `distribute` which does not run automatically like `prune`. will not hurt, but the gateway won't be moved to `leaving` until the next `Tick` comes in

src/gar.lua
```diff
@@ -96,6 +96,9 @@ GatewayRegistrySettings = GatewayRegistrySettings
 		},
 	}
 
+--- @type Timestamp|nil
+NextGatewaysPruneTimestamp = NextGatewaysPruneTimestamp or 0
```
fedellen, line 100: whats the main difference between 0 and nil here for the logic? 

```suggestion
--- @type Timestamp
NextGatewaysPruneTimestamp = NextGatewaysPruneTimestamp or 0
```

src/gar.lua
```diff
@@ -895,12 +907,23 @@ function gar.pruneGateways(currentTimestamp, msgId)
 		delegateStakeWithdrawing = 0,
 		stakeSlashed = 0,
 	}
+	if not NextGatewaysPruneTimestamp or currentTimestamp < NextGatewaysPruneTimestamp then
+		-- No known pruning work to do
+		return result
+	end
```
fedellen, line 913: ah, 0 must be truthy.

fedellen (approved):

src/gar.lua
arielmelendez: Good to know. Will remove to avoid complexity.

src/gar.lua
arielmelendez: It's an initial condition in this case. When we deploy updates to the source, globals will have previous values available to assign here. 0 indicates "definitely prune next time pruning happens". `nil` indicates that there's nothing to prune. 0 is more conservative when we're not sure what the deal should be.

src/gar.lua
arielmelendez: Correct, 0 is truthy in Lua

src/gar.lua
arielmelendez: 629ba1f

PR #218, opened 2024-11-21
feat(redelegate stake): slim down target and source gateway response

fedellen:


spec/gar_spec.lua
```diff
@@ -2940,6 +2940,11 @@ describe("gar", function()
 			}
 			targetGateway.totalDelegatedStake = qty
 
+			sourceGateway.delegates = nil
+			sourceGateway.vaults = nil
+			targetGateway.delegates = nil
+			targetGateway.vaults = nil
+
```
fedellen, line 2947: nit: we're losing coverage here. I believe we can instead keep the expectations but use _G

src/gar.lua
```diff
@@ -1848,8 +1848,8 @@ function gar.redelegateStake(params)
 	GatewayRegistry[targetAddress] = targetGateway
 
 	return {
-		sourceGateway = sourceGateway,
-		targetGateway = targetGateway,
+		sourceGateway = utils.deepCopy(sourceGateway, { "delegates", "vaults", "settings.allowedDelegatesLookup" }),
+		targetGateway = utils.deepCopy(targetGateway, { "delegates", "vaults", "settings.allowedDelegatesLookup" }),
 		redelegationFee = redelegationFee,
 		feeResetTimestamp = currentTimestamp + constants.redelegationFeeResetIntervalMs,
 		redelegationsSinceFeeReset = Redelegations[delegateAddress].redelegations,
```
fedellen, line 1855: nit: can share state above. `local redelegationsSinceFeeReset = (previousRedelegations and previousRedelegations.redelegations or 0) + 1` onto ~L1838

src/gar.lua
```diff
@@ -1836,23 +1836,22 @@ function gar.redelegateStake(params)
...
+	Redelegations[delegateAddress] = {
+		timestamp = currentTimestamp,
+		redelegations = redelegationsSinceFeeReset,
+	}
 
 	return {
-		sourceGateway = sourceGateway,
-		targetGateway = targetGateway,
+		sourceAddress = sourceAddress,
+		targetAddress = targetAddress,
```
dtfiedler, line 1851: i was suggesting, just changing this in the Handler result 😄 , but this is fine too

dtfiedler (approved):

PR #228, opened 2024-11-25
PE-7147: Prune Epochs Only When Necessary

arielmelendez:
This one is a little different than the others because there's a very structured way in which the epochs are created and pruned. There's no need to reset the next prune timestamp after user actions here.

Note: I modified the deepCopy behavior to:
- do a very fast, basic deep copy if no exclusions are present
- if exclusions are present, activate the reindexing feature for numeric Arrays (not dictionaries)

### DIFF
--- a/src/epochs.lua
+++ b/src/epochs.lua
@@ -17,7 +17,7 @@ local epochs = {}
 --- @field distributions Distribution The distributions of the epoch
 
 --- @class EpochSettings
---- @field maxCachedEpochsCount number The number of epochs to prune
+--- @field pruneEpochsCount number The number of epochs to prune
 --- @field prescribedNameCount number The number of prescribed names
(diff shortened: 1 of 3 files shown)


src/epochs.lua
```diff
@@ -61,7 +61,7 @@ local epochs = {}
 Epochs = Epochs or {}
 EpochSettings = EpochSettings
 	or {
-		pruneEpochsCount = 14, -- prune epochs older than 14 days
+		maxCachedEpochsCount = 14, -- prune epochs older than 14 days
```
dtfiedler, line 64: note: this will require manually setting this value in existing processes and removing the existing settings, given our use of `or`

arielmelendez: 60bcdc8

dtfiedler (approved):

PR #229, opened 2024-11-26
fix(gar): update gar settings, minimum stakes, withdrawal lengths

dtfiedler:
Moves the gar functions to helpers.mjs, and updates tests with new values

Includes PE-7186/PE-7187/PE-7188

### DIFF
--- a/src/gar.lua
+++ b/src/gar.lua
@@ -75,26 +75,46 @@ local gar = {}
 --- @type Gateways
 GatewayRegistry = GatewayRegistry or {}
 
-GatewayRegistrySettings = GatewayRegistrySettings
-	or {
-		observers = {
-			maxPerEpoch = 50,
-			tenureWeightDays = 180,
-			tenureWeightPeriod = 180 * 24 * 60 * 60 * 1000, -- aproximately 180 days
-			maxTenureWeight = 4,
-		},
-		operators = {
-			minStake = 50000 * 1000000, -- 50,000 IO
-			withdrawLengthMs = 30 * 24 * 60 * 60 * 1000, -- 30 days to lower operator stake
-			leaveLengthMs = 90 * 24 * 60 * 60 * 1000, -- 90 days that balance will be vaulted
-			failedEpochCountMax = 30, -- number of epochs failed before marked as leaving
-			failedEpochSlashPercentage = 0.2, -- 20% of stake is returned to protocol balance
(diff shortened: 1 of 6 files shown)


src/gar.lua
```diff
@@ -75,26 +75,46 @@ local gar = {}
...
+
+--- @class DelegateSettings
+--- @field minStake number
+--- @field withdrawLengthMs number
+
+--- @class GatewayRegistrySettings
+--- @field observers ObserverSettings
+--- @field operators OperatorSettings
+--- @field delegates DelegateSettings
+GatewayRegistrySettings = {
```
dtfiedler, line 99: removed the `or` here as these are now consistent across both testnet and devnet

tests/helpers.mjs
dtfiedler, line 1: moved the handlers from `gar.test.mjs` to here for re-use

src/gar.lua
```diff
@@ -75,26 +75,46 @@ local gar = {}
...
+	observers = {
+		maxPerEpoch = 50,
+		tenureWeightDays = 180,
+		tenureWeightPeriod = 180 * 24 * 60 * 60 * 1000, -- approximately 180 days
+		maxTenureWeight = 4,
+	},
+	operators = {
+		minStake = 10000 * 1000000, -- 10,000 IO
```
arielmelendez, line 107: Nit: would feel a bit more comfortable to have 1000000 everywhere in the code set to a constant for mIOPerIO or something like that. It can be tricky to count the zeros visually beyond a small few.

src/gar.lua
```diff
@@ -75,26 +75,46 @@ local gar = {}
...
+		withdrawLengthMs = 90 * 24 * 60 * 60 * 1000, -- 90 days to lower operator stake
+		leaveLengthMs = 90 * 24 * 60 * 60 * 1000, -- 90 days that balance will be vaulted
+		failedEpochCountMax = 30, -- number of epochs failed before marked as leaving
+		failedEpochSlashRate = 0.2, -- 20% of stake is returned to protocol balance
```
arielmelendez, line 111: Nice name update.

arielmelendez (approved):

src/gar.lua
dtfiedler: agree - i'll create.a ticket to standardrize all these

PR #246, opened 2024-12-04
PE-7219: FYEO feedback round 1

arielmelendez:


spec/epochs_spec.lua
```diff
@@ -159,24 +159,27 @@ describe("epochs", function()
...
 
-	describe("computePrescrbiedNamesForEpoch", function()
+	describe("computePrescribedNamesForEpoch", function()
+		-- NOTE: Record names in the tests below use spelled out numbers because without that
+		-- there's insufficient base64url information encoded in the final encoded block to
+		-- disambiguate the decoded vallues.
 		it("should return all eligible names if fewer than the maximum in name registry", function()
 			_G.NameRegistry.records = {
-				["arns-name-1"] = {
+				["arns-name-one"] = {
```
fedellen, line 168: So is this saying that `arns-name-1` and `arns-name-2` are the same base64? How would one decode that base64 back to `arns-name-1` or 2 accurately?

fedellen: Perhaps they just need to be a power of 4 since these are supposed to be b64 values but we're stubbing them as strings?

arielmelendez: The problem is that we're passing strings that are NOT base64 encodings of anything in a base64 decoder so certain characters get treated as padding and are thrown out. We use the decoded values to then sort a dictionary. The dictionary's iteration order is undefined by Lua, so if you end up with a bunch of decoded values that are the same, the dictionary will sort unstably into an array.

PR #250, opened 2024-12-06
test(save-observation): init integration coverage of prescribed observations PE-7210

fedellen:


tests/helpers.mjs
```diff
@@ -83,15 +95,15 @@ export const transfer = async ({
     },
     memory,
   );
-  assertNoResultError(transferResult);
```
arielmelendez, line 86: Did you intend to parameterize the error checking for this rather than simply removing here?

arielmelendez: OH i see true is the default for handle.

tests/helpers.mjs
```diff
@@ -16,18 +16,30 @@ const initialOperatorStake = 100_000_000_000;
...
+export const genesisEpochTimestamp = 1719900000000; // Tuesday, July 2, 2024, 06:00:00 AM UTC
+export const epochLength = 1000 * 60 * 60 * 24; // 24 hours
+export const distributionDelay = 1000 * 60 * 40; // 40 minutes
+
 const { handle: originalHandle, memory } = await createAosLoader();
 export const startMemory = memory;
 
-export async function handle(options = {}, mem = startMemory) {
-  return originalHandle(
+export async function handle(
```
arielmelendez, line 26: Was it going to be too big a lift to move to an object for args? All the `, undefined, false)` that's starting to appear in the code is a bit unsightly...

tests/gar.test.mjs
```diff
@@ -1200,10 +1204,160 @@ describe('GatewayRegistry', async () => {
...
+            delegateRewards: [],
+          },
+        },
+      });
+
+      // Assert prescribed observers
+      const prescribedObservers = JSON.parse(futureTick.Messages[0].Data)
+        .maybeNewEpoch.prescribedObservers;
+      assert.equal(prescribedObservers.length, 2);
+      assert.equal(prescribedObservers[0].observerAddress, STUB_ADDRESS);
```
arielmelendez, line 1260: Will these sort stably all the time?

arielmelendez (approved):
Looks like a good start. Might want to assess how big a lift it is to move to an object argument for handle rather than more positional params.

tests/gar.test.mjs
fedellen: good question. possibly not

tests/helpers.mjs
fedellen: yeah it seemed large, but should be doable -- I'll timebox it

tests/gar.test.mjs
fedellen: Fixed in the refactor, they were indeed sorted randomly: [0b61e65](https://github.com/ar-io/ar-io-network-process/pull/250/commits/0b61e65340fcd055492b9d96d2a7bbd781198be0)

tests/helpers.mjs
fedellen: [0b61e65](https://github.com/ar-io/ar-io-network-process/pull/250/commits/0b61e65340fcd055492b9d96d2a7bbd781198be0)

PR #252, opened 2024-12-06
feat(primary name pricing): init change to 1 undername price PE-7230

fedellen:


tests/arns.test.mjs
```diff
@@ -543,16 +544,28 @@ describe('ArNS', async () => {
...
+            { name: 'Intent', value: 'Primary-Name-Request' },
+            { name: 'Name', value: 'test-name' },
+          ],
+        },
+        buyMemory,
+      );
       assertNoResultError(result);
       const tokenCost = JSON.parse(result.Messages[0].Data);
-      assert.equal(tokenCost, 10000000);
+      assert.equal(tokenCost, 500000);
```
arielmelendez, line 568: Could go further here to assert that it's the same cost as 1 undername.

arielmelendez (dismissed):

PR #258, opened 2024-12-09
PE-7224: Rename tIO, mIO, IO to tARIO, mARIO, and ARIO

arielmelendez:
Did NOT change references to ioEvent to arioEvent as it would have nuked the diff on main.lua. Can do it another time if necessary.


src/ario_event.lua
```diff
@@ -1,8 +1,9 @@
 local AOEvent = require("ao_event")
 local utils = require("utils")
 
+-- TODO: RENAME TO ARIOEvent
 -- Convenience factory function for prepopulating analytic and msg fields into AOEvents
-local function IOEvent(msg, initialData)
+local function ARIOEvent(msg, initialData)
```
fedellen, line 6: looks to be done

fedellen (approved):

src/ario_event.lua
arielmelendez: Didn't rename the ioEvent variable on messages though.

PR #264, opened 2024-12-11
PE-7126: primary name events

arielmelendez:


tests/helpers.mjs
```diff
@@ -40,6 +40,22 @@ export function assertNoResultError(result) {
   assert.strictEqual(errorTag, undefined);
 }
 
+export function parseEventsFromResult(result) {
```
fedellen, line 43: nice

tests/primary.test.mjs
```diff
@@ -225,6 +249,29 @@ describe('primary names', function () {
       });
 
     assertNoResultError(approvePrimaryNameRequestResult);
+    const parsedApproveEvents = parseEventsFromResult(
+      approvePrimaryNameRequestResult,
+    );
+    assert.equal(parsedApproveEvents.length, 1);
+    assert.deepStrictEqual(parsedApproveEvents[0], {
+      _e: 1,
```
fedellen, line 257: is `_e` always 1 or does it correspond to length?

fedellen (dismissed):

tests/primary.test.mjs
arielmelendez: It's always 1. It's just a flag field for the parser on the CU side.

fedellen (approved):

PR #266, opened 2024-12-11
refactor: use correct `stakes `syntax in fundFrom assertion message

fedellen:


atticusofsparta (approved):

arielmelendez (approved):
Bonus points if you DRY that out :)

PR #272, opened 2024-12-13
feat(distribution rate): init time based distribution reward rate PE-7190

fedellen:


src/epochs.lua
```diff
@@ -766,4 +768,18 @@ function epochs.nextEpochsPruneTimestamp()
 	return NextEpochsPruneTimestamp
 end
 
+-- Swap out the use of the hard-coded value to use a time-based function with the above criteria.
```
fedellen, line 771: leftover comment
```suggestion
```

tests/tick.test.mjs
```diff
@@ -415,8 +415,8 @@ describe('Tick', async () => {
     });
 
     // get the epoch timestamp and assert it is in 24 hours
-    const protocolBalanceAtStartOfEpoch = 50_000_000_0000; // 50M ARIO
-    const totalEligibleRewards = protocolBalanceAtStartOfEpoch * 0.05; // 5% of the protocol balance
+    const protocolBalanceAtStartOfEpoch = 50_000_000_000_000; // 50M ARIO
+    const totalEligibleRewards = protocolBalanceAtStartOfEpoch * 0.001; // 0.1% of the protocol balance
```
fedellen, line 419: this expectation was actually 0.05% of the balance, but the 50M ARIO value was wrong. so the `// 50M ARIO` comment was correct, but the `//  0.1% of of the protocol balance` and the mARIO number value was incorrect

tests/tick.test.mjs
```diff
@@ -584,11 +584,11 @@ describe('Tick', async () => {
         totalEpochCount: 1,
       },
       weights: {
-        compositeWeight: 22,
+        compositeWeight: 32,
         gatewayRewardRatioWeight: 1,
         normalizedCompositeWeight: 1,
         observerRewardRatioWeight: 1,
-        stakeWeight: 5.5,
+        stakeWeight: 8,
```
fedellen, line 591: I haven't grasped why this reward rate change effects weights yet

tests/tick.test.mjs
```diff
@@ -415,8 +415,8 @@ describe('Tick', async () => {
     });
 
     // get the epoch timestamp and assert it is in 24 hours
-    const protocolBalanceAtStartOfEpoch = 50_000_000_0000; // 50M ARIO
```
fedellen, line 418: this expectations were actually 0.05% of the balance, but the 50M ARIO value was wrong. so the // 50M ARIO comment was correct, but the //  5% of of the protocol balance and the mARIO number value was incorrect

spec/epochs_spec.lua
```diff
@@ -937,4 +938,30 @@ describe("epochs", function()
 			assert.are.equal(currentTimestamp + 1, _G.NextEpochsPruneTimestamp)
 		end)
 	end)
+
+	describe("getRewardRateForEpoch", function()
+		it("returns 0.05% for the first 365 epochs (one year)", function()
+			assert.are.equal(0.0005, epochs.getRewardRateForEpoch(0))
+			assert.are.equal(0.0005, epochs.getRewardRateForEpoch(1))
+			assert.are.equal(0.0005, epochs.getRewardRateForEpoch(364))
+			assert.are.equal(0.0005, epochs.getRewardRateForEpoch(365))
```
arielmelendez, line 947: If we're zero-indexing epochs, shouldn't the final epoch at 0.05% be 364?

src/constants.lua
```diff
@@ -11,6 +11,10 @@ constants.oneYearMs = 31536000 * 1000
 
 -- EPOCHS
 constants.defaultEpochDurationMs = constants.oneDayMs
+constants.decayStartingRewardRate = 0.001
+constants.defaultRewardRate = 0.0005
+constants.rewardDecayStartEpoch = 365
+constants.rewardDecayEndEpoch = 548
```
arielmelendez, line 17: `End` in the name is squishy because it's not clear whether it's inclusive or exclusive. Perhaps `Last` is clearer (and uses an inclusive value)?

src/epochs.lua
```diff
@@ -766,4 +768,17 @@ function epochs.nextEpochsPruneTimestamp()
 	return NextEpochsPruneTimestamp
 end
 
+function epochs.getRewardRateForEpoch(epochIndex)
+	if epochIndex <= constants.rewardDecayStartEpoch then
+		return constants.defaultRewardRate
+	elseif epochIndex < constants.rewardDecayEndEpoch then
+		return constants.decayStartingRewardRate
+			- (epochIndex - constants.rewardDecayStartEpoch)
+				* (constants.decayStartingRewardRate - constants.defaultRewardRate)
```
arielmelendez, line 777: Could improve readability here with a local variable like `perEpochDecayRate`.

arielmelendez (approved):
Left some nits on readability improvements. Thanks!

src/constants.lua
```diff
@@ -11,6 +11,10 @@ constants.oneYearMs = 31536000 * 1000
 
 -- EPOCHS
 constants.defaultEpochDurationMs = constants.oneDayMs
+constants.decayStartingRewardRate = 0.001
```
dtfiedler, line 14:
```suggestion
constants.maxiumRewardRate = 0.001
constants.minimumRewardRate = 0.0005
```
and then use these when computing the current decay rate

spec/epochs_spec.lua
fedellen: I think the first epoch is EpochIndex 1. so the test is confusing to check for 0, removed that

src/epochs.lua
fedellen: [0d8533f](https://github.com/ar-io/ar-io-network-process/pull/272/commits/0d8533f23cc03f72ce61f34f1e41319b1b7ef2bd)

src/constants.lua
fedellen: [0d8533f](https://github.com/ar-io/ar-io-network-process/pull/272/commits/0d8533f23cc03f72ce61f34f1e41319b1b7ef2bd)

dtfiedler (approved):

PR #273, opened 2024-12-13
feat(primary names): add processID to read APIs PE-7307

fedellen:


src/primary_names.lua
```diff
@@ -321,10 +338,19 @@ function primaryNames.getPaginatedPrimaryNames(cursor, limit, sortBy, sortOrder)
 	local primaryNamesArray = {}
 	local cursorField = "name"
 	for owner, primaryName in pairs(primaryNames.getUnsafePrimaryNameOwners()) do
+		if not primaryName.processId then
+			-- TODO: After initial deployment, we can make processId not nil and remove this block
+			local record = arns.getRecord(baseNameForName(primaryName.name))
+			if record then
+				primaryName.processId = record.processId
+				PrimaryNames.owners[owner] = primaryName
```
dtfiedler, line 346: rather than storing this in state, we can compute it on read using the requested name to get the base name

src/primary_names.lua
```diff
@@ -237,29 +247,39 @@ end
...
 	end
+
+	if not nameData.processId then
+		-- TODO: After initial deployment, we can make processId not nil and remove this block
+		local record = arns.getRecord(baseNameForName(nameData.name))
+		if record then
+			nameData.processId = record.processId
+			PrimaryNames.owners[address] = nameData
+		end
+	end
```
dtfiedler, line 271: nit - this isn't something that needs to be stored in state, it's cheap to compute on the fly and return with the name data

src/primary_names.lua
```diff
@@ -237,29 +247,39 @@ end
...
 	return {
 		owner = address,
 		name = nameData.name,
 		startTimestamp = nameData.startTimestamp,
+		processId = nameData.processId,
```
dtfiedler, line 276:
```suggestion
	local record = arns.getRecord(baseNameForName(nameData.name))
	return {
		owner = address,
		name = nameData.name,
		startTimestamp = nameData.startTimestamp,
		processId = record.processId,
```

src/primary_names.lua
```diff
@@ -321,10 +338,19 @@ function primaryNames.getPaginatedPrimaryNames(cursor, limit, sortBy, sortOrder)
...
+			if record then
+				primaryName.processId = record.processId
+				PrimaryNames.owners[owner] = primaryName
+			end
+		end
 		table.insert(primaryNamesArray, {
 			name = primaryName.name,
 			owner = owner,
 			startTimestamp = primaryName.startTimestamp,
+			processId = primaryName.processId,
```
dtfiedler, line 353:
```suggestion
		local record = arns.getRecord(baseNameForName(primaryName.name))
		table.insert(primaryNamesArray, {
			name = primaryName.name,
			owner = owner,
			startTimestamp = primaryName.startTimestamp,
			processId = record.processId,
```

src/primary_names.lua
fedellen: this seemed like a lot of getRecord calls when dealing with a paginated getPrimaryNames call returning 1000 names for example

src/primary_names.lua
```diff
@@ -237,29 +243,34 @@ end
...
 --- @param address string
---- @return PrimaryNameWithOwner|nil primaryNameWithOwner - the primary name with owner data, or nil if it does not exist
+--- @return PrimaryNameInfo|nil -- the primary name with owner data, or nil if it does not exist
 function primaryNames.getPrimaryNameDataWithOwnerFromAddress(address)
 	local nameData = PrimaryNames.owners[address]
+
 	if not nameData then
 		return nil
 	end
+	local name = baseNameForName(nameData.name)
```
dtfiedler, line 260: nit: use `baseName` to avoid confusing with nameData.name

dtfiedler (approved):
we'll want to update the SDK type with this new attribute

src/primary_names.lua
dtfiedler: fetching records is cheap (o(1)) since they are indexed by name. we could memoize the process id for a given base name if we wanted to get really efficient

src/primary_names.lua
fedellen: ah, I'll condense this. was breaking it out for debug logging

src/primary_names.lua
dtfiedler: thinking further - memoizing wouldn't give us any improvements, and actually increase the memory footprint - disregard

PR #279, opened 2024-12-17
feat(gar): optimize prescribed observers, use lookup table and comput…

dtfiedler:
…e full array on read handlers

### DIFF
--- a/src/main.lua
+++ b/src/main.lua
@@ -1946,10 +1946,31 @@ addEventingHandler(
 
 		local epochIndex = providedEpochIndex or epochs.getEpochIndexForTimestamp(timestamp)
 		local prescribedObservers = epochs.getPrescribedObserversForEpoch(epochIndex)
+		-- Iterate over prescribed observers and add gateway details
+		local prescribedObserversWithWeights = {}
+		for _, gatewayAddress in ipairs(prescribedObservers) do
+			local gateway = gar.getGateway(gatewayAddress)
+			if gateway then
+				table.insert(prescribedObserversWithWeights, {
+					observerAddress = gateway.observerAddress,
+					gatewayAddress = gatewayAddress,
+					normalizedCompositeWeight = gateway.weights.normalizedCompositeWeight,
+					stakeWeight = gateway.weights.stakeWeight,
+					tenureWeight = gateway.weights.tenureWeight,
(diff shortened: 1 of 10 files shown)


src/main.lua
```diff
@@ -597,6 +597,12 @@ addEventingHandler(ActionMap.Transfer, utils.hasMatchingTag("Action", ActionMap.
 		msg.ioEvent:addField("RecipientNewBalance", recipientNewBalance)
 	end
 
+	-- if the sender is the protocol, then we need to update the circulating supply as tokens are now in circulation
+	if msg.From == ao.id then
+		LastKnownCirculatingSupply = LastKnownCirculatingSupply + quantity
+		addSupplyData(msg.ioEvent)
+	end
```
fedellen, line 604: hmm, saw some test errors when looking into that `beforeEAch` mystery that might be related to this fix

src/main.lua
```diff
@@ -1903,12 +1909,8 @@ end)
 
 addEventingHandler(ActionMap.Epoch, utils.hasMatchingTag("Action", ActionMap.Epoch), function(msg)
 	-- check if the epoch number is provided, if not get the epoch number from the timestamp
-	local providedEpochIndex = msg.Tags["Epoch-Index"]
-	local timestamp = msg.Timestamp
-
-	assert(providedEpochIndex or timestamp, "Epoch index or timestamp is required")
-
-	local epochIndex = providedEpochIndex or epochs.getEpochIndexForTimestamp(timestamp)
+	local epochIndex = msg.Tags["Epoch-Index"] and tonumber(msg.Tags["Epoch-Index"])
```
fedellen, line 1912: is epoch-index already sanitized as number?

fedellen (commented):
Looks good -- I think its ready but needs re-integration with latest develop

src/main.lua
dtfiedler: https://github.com/ar-io/ar-io-network-process/blob/develop/src/utils.lua#L676

fedellen (dismissed):

dtfiedler:
just FYI - it'll take 14 epochs to prune out all the old state of previous epochs. As mentioned, this type is not used externally anywhere so no breaking change.

fedellen (approved):

PR #283, opened 2024-12-18
fix(gar): do not remove delegate before decreasing stake

dtfiedler:


src/main.lua
```diff
@@ -395,7 +395,6 @@ local function assertAndSanitizeInputs(msg)
 			.. msg.Timestamp
 	)
 	assert(msg.From, "From is required")
-	assert(msg.Id, "Id is required")
```
dtfiedler, line 398: this was blocking some dry-runs like `Info` - but not others 🤔

src/gar.lua
```diff
@@ -2055,7 +2067,6 @@ function unlockGatewayDelegateVault(gateway, delegateAddress, vaultId)
 	assert(vault, "Vault not found")
 
 	balances.increaseBalance(delegateAddress, vault.balance)
-	gateway.delegates[delegateAddress] = nil
```
dtfiedler, line 2058: we nil out delegate in `decreaseDelegateStakeAtGateway` if necessary

dtfiedler: i'm also not sure if we should be calling `decreaseDelegateStakeAtGateway` here

dtfiedler: [7b55fd5](https://github.com/ar-io/ar-io-network-process/pull/283/commits/7b55fd5ee23add7948d21d23b07ccc393745d35c)

src/gar.lua
```diff
@@ -489,8 +496,15 @@ end
 --- @param ban boolean|nil do not add the delegate back to the gateway allowlist if their delegation is over
 function decreaseDelegateStakeAtGateway(delegateAddress, gateway, quantity, ban)
 	local delegate = gateway.delegates[delegateAddress]
-	-- use this in an inverse way
-	increaseDelegateStakeAtGateway(delegate, gateway, -quantity)
+	assert(delegate, "Delegate is required")
+	-- zero is allowed as it is a no-op
+	assert(
+		quantity and utils.isInteger(quantity) and quantity >= 0,
```
dtfiedler, line 502: nit: we could change this to be greater than the delegate.totalDelegatedStake

fedellen (approved):

PR #288, opened 2024-12-18
fix(balance): fix tag check for balance (#285)

dtfiedler:


fedellen (approved):

src/main.lua
```diff
@@ -1851,17 +1851,17 @@ addEventingHandler(ActionMap.Balances, Handlers.utils.hasMatchingTag("Action", A
...
 addEventingHandler(ActionMap.Balance, Handlers.utils.hasMatchingTag("Action", ActionMap.Balance), function(msg)
-	local target = msg.Tags.Target or msg.Tags.Address or msg.From
+	local target = msg.Tags.Target or msg.Tags.Address or msg.Tags.Recipient or msg.From
 	local balance = balances.getBalance(target)
 
 	-- must adhere to token.lua spec for arconnect compatibility
 	Send(msg, {
 		Target = msg.From,
 		Action = "Balance-Notice",
+		Account = msg.Tags.Address or msg.Tags.Recipient or msg.From,
```
arielmelendez, line 1861: Why not just use target here?

dtfiedler: just going off of the token.lua spec - not asking questions

PR #293, opened 2024-12-19
feat: audit msg.Timestamp, remove input of msg.Tags.Timestamp PE-7338

fedellen:


src/main.lua
```diff
@@ -409,6 +408,17 @@ local function updateLastKnownMessage(msg)
...
+--- @field Id string
+--- @field From string
+--- @field Timestamp number
+--- @field Tags table<string, any>
+
+--- @param handlerName string
+--- @param pattern fun():string
+--- @param handleFn fun(msg: ParsedMessage)
+--- @param critical boolean?
+--- @param printEvent boolean?
```
fedellen, line 421: gives us IDE types in main.handlers
image

fedellen: will include ioEvent

dtfiedler (dismissed):

src/main.lua
```diff
@@ -553,7 +569,7 @@ end, function(msg)
 		addSupplyData(msg.ioEvent)
 	end
 
-	return prunedStateResult
+	-- return prunedStateResult -- TODO: need to return?
```
fedellen, line 572: appears tests work fine without returning here in prune. its the only function returning so the lua type annotations was calling it out -- will remove

PR #317, opened 2025-01-06
feat(distribute): send an `Epoch-Distribution-Notice` when epochs are…

dtfiedler:
… distributed

### DIFF
--- a/src/epochs.lua
+++ b/src/epochs.lua
@@ -525,11 +525,14 @@ function epochs.computeTotalEligibleRewardsForEpoch(epochIndex, prescribedObserv
 	local protocolBalance = balances.getBalance(ao.id)
 	local rewardRate = epochs.getRewardRateForEpoch(epochIndex)
 	local totalEligibleRewards = math.floor(protocolBalance * rewardRate)
-	local eligibleGatewayReward =
-		math.floor(totalEligibleRewards * constants.gatewayOperatorRewardRatio / #activeGatewayAddresses)
-	local eligibleObserverReward = math.floor(
-		totalEligibleRewards * constants.observerRewardRatio / utils.lengthOfTable(prescribedObserversLookup)
-	)
+	local eligibleGatewayReward = #activeGatewayAddresses > 0
+			and math.floor(totalEligibleRewards * constants.gatewayOperatorRewardRatio / #activeGatewayAddresses)
+		or 0
+	local eligibleObserverReward = utils.lengthOfTable(prescribedObserversLookup) > 0
+			and math.floor(
+				totalEligibleRewards * constants.observerRewardRatio / utils.lengthOfTable(prescribedObserversLookup)
+			)
+		or 0
 	-- compute for each gateway what their potential rewards are and for their delegates
 	local potentialRewards = {}
 	-- use ipairs as activeGatewayAddresses is an array
--- a/src/main.lua
+++ b/src/main.lua
@@ -1667,7 +1667,9 @@ end)
 
 -- distribute rewards
 -- NOTE: THIS IS A CRITICAL HANDLER AND WILL DISCARD THE MEMORY ON ERROR
-addEventingHandler("distribute", utils.hasMatchingTag("Action", "Tick"), function(msg)
+addEventingHandler("distribute", function(msg)
+	return msg.Action == "Tick" or msg.Action == "Distribute"
+end, function(msg)
 	local msgId = msg.Id
 	local blockHeight = tonumber(msg["Block-Height"])
 	local hashchain = msg["Hash-Chain"]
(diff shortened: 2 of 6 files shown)


tests/tick.test.mjs
```diff
@@ -455,8 +477,29 @@ describe('Tick', async () => {
       timestamp: distributionTimestamp,
     });
 
-    // assert no error tag
-    assertNoResultError(distributionTick);
+    // assert multiple messages are sent given the tick notice, epoch created notice and epoch distribution notice
+    assert.equal(distributionTick.result.Messages.length, 4); // cannot explain why this is 4, i'd expect it to be 3 messages (1 tick notice, 1 epoch created notice, 1 epoch distribution notice)
```
dtfiedler, line 481: added a comment here - we are getting one extra `Tick-Notice` at the end of the message array, unclear why. will spend some more time on it

fedellen: Whats the 4th message?

fedellen (dismissed):

src/epochs.lua
```diff
@@ -525,11 +525,14 @@ function epochs.computeTotalEligibleRewardsForEpoch(epochIndex, prescribedObserv
...
-		totalEligibleRewards * constants.observerRewardRatio / utils.lengthOfTable(prescribedObserversLookup)
-	)
+	local eligibleGatewayReward = #activeGatewayAddresses > 0
+			and math.floor(totalEligibleRewards * constants.gatewayOperatorRewardRatio / #activeGatewayAddresses)
+		or 0
+	local eligibleObserverReward = utils.lengthOfTable(prescribedObserversLookup) > 0
+			and math.floor(
+				totalEligibleRewards * constants.observerRewardRatio / utils.lengthOfTable(prescribedObserversLookup)
+			)
+		or 0
```
dtfiedler, line 535: this was causing `inf` due to rounding by 0 when there are no gateways joined

src/main.lua
```diff
@@ -1701,41 +1714,56 @@ addEventingHandler("distribute", utils.hasMatchingTag("Action", "Tick"), functio
 	local tickedRewardDistributions = {}
 	local totalTickedRewardsDistributed = 0
 
+	print("Ticking from " .. lastTickedEpochIndex .. " to " .. targetCurrentEpochIndex)
+
 	--- tick to the newest epoch, and stub out any epochs that are not yet created
 	for i = lastTickedEpochIndex + 1, targetCurrentEpochIndex do
+		print("Ticking epoch " .. i)
```
dtfiedler, line 1721: these logs proved useful when debuggin tests, so left them in

src/main.lua
```diff
@@ -1701,41 +1714,56 @@ addEventingHandler("distribute", utils.hasMatchingTag("Action", "Tick"), functio
...
 		-- and that we don't try and distribute before an epoch is created
 		local tickTimestamp = math.min(msg.Timestamp, epochDistributionTimestamp)
 		-- TODO: if we need to "recover" epochs, we can't rely on just the current message hashchain and block height,
 		-- we should set the prescribed observers and names to empty arrays and distribute rewards accordingly
 		local tickResult = tick.tickEpoch(tickTimestamp, blockHeight, hashchain, msgId)
-		if tickTimestamp == epochDistributionTimestamp then
-			-- if we are distributing rewards, we should update the last ticked epoch index to the current epoch index
-			LastTickedEpochIndex = i
-			table.insert(tickedEpochIndexes, i)
-		end
```
dtfiedler, line 1717: this never made sense - if we've ticked the epoch at all, we should increment the `LastTickedEpochIndex`. We can also consider replacing it with two more granular global variables `LastCreatedEpochIndex` and `LastDistributedEpochIndex`

tests/tick.test.mjs
```diff
@@ -455,8 +474,31 @@ describe('Tick', async () => {
...
+      ).value,
+      'Epoch-Distribution-Notice',
+    );
+
+    assert.equal(
+      distributionTick.result.Messages[2].Tags.find(
+        (tag) => tag.name === 'Action',
+      ).value,
+      'Tick-Notice',
+    );
```
dtfiedler, line 501: validates all three messages get sent when a new epoch is created and distributed

tests/tick.test.mjs
dtfiedler, line 1: most of the changes in this file are to leverage the reusable `tick` handler in `helpers.mjs` and update state with correct variables from the contract

tests/tick.test.mjs
dtfiedler: it was mainly due to test setup, and some minor edge cases in how we create the first epoch. The latest commits have a few changes that address the issue.

src/main.lua
dtfiedler: we could also add `LastDemandFactorUpdateTimestamp` to track when the demand factor was last updated.

fedellen (approved):

src/main.lua
dtfiedler: merged into his branch on https://github.com/ar-io/ar-io-network-process/pull/317/commits/49661f138b4e8a7d8a75dfe4cbceb76526853f7c

src/main.lua
```diff
@@ -1667,79 +1681,84 @@ end)
...
-		Send(msg, {
-			Target = msg.From,
-			Action = "Tick-Notice",
-			LastTickedEpochIndex = tostring(LastTickedEpochIndex),
-			Data = json.encode(tickResult),
-		})
+	for i = lastCreatedEpochIndex, targetCurrentEpochIndex do
```
dtfiedler, line 1722: this logically makes more sense - instead of "going back one epoch", start this process at the last known created epoch and go until we reach the target current epoch

PR #321, opened 2025-01-09
feat: add all paginated delegates to main handlers PE-7221

fedellen:
this PR adds a handler that returns paginated delegates from all gateways in the shape of 

```ts
{
  address: string, // Address of the delegate
  gatewayAddress: string, // Gateway address the stake is delegated to
  delegatedStake: number, // Amount of stake currently delegated to gateway
  startTimestamp: number, // Timestamp of when the delegate first staked to this gateway
  vaultedStake: number, // Aggregation of all stake found in any vaults found for the delegate (stake being withdrawn)
},
```

### DIFF
--- a/src/gar.lua
+++ b/src/gar.lua
@@ -2149,4 +2149,44 @@ function gar.nextRedelegationsPruneTimestamp()
 	return NextRedelegationsPruneTimestamp
 end
 
+--- @class DelegatesFromAllGateways
+--- @field address WalletAddress
+--- @field gatewayAddress WalletAddress
+--- @field startTimestamp Timestamp
+--- @field delegatedStake mARIO
+--- @field vaultedStake mARIO
+
+--- @param cursor string|nil
+--- @param limit number
+--- @param sortBy string|nil
+--- @param sortOrder string|nil
+--- @return PaginatedTable<DelegatesFromAllGateways>
+function gar.getPaginatedDelegatesFromAllGateways(cursor, limit, sortBy, sortOrder)
+	--- @type DelegatesFromAllGateways[]
+	local allDelegations = {}
+
+	for gatewayAddress, gateway in pairs(GatewayRegistry) do
+		for delegateAddress, delegate in pairs(gateway.delegates) do
+			table.insert(allDelegations, {
+				address = delegateAddress,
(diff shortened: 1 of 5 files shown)


tests/gar.test.mjs
```diff
@@ -1424,6 +1424,63 @@ describe('GatewayRegistry', async () => {
...
+              { name: 'Limit', value: '1' },
+              { name: 'Sort-By', value: 'delegatedStake' },
+              { name: 'Sort-Order', value: 'asc' },
+            ],
+          },
+          memory: secondDelegatedStakeMemory,
+          timestamp: STUB_TIMESTAMP + 2,
+        });
+        // parse items, nextCursor
+        console.log('paginatedDelegates', paginatedDelegates);
```
fedellen, line 1464:
```suggestion
```

src/gar.lua
```diff
@@ -2191,4 +2191,46 @@ function gar.nextRedelegationsPruneTimestamp()
...
+				address = delegateAddress,
+				gatewayAddress = gatewayAddress,
+				startTimestamp = delegate.startTimestamp,
+				delegatedStake = delegate.delegatedStake,
+				vaultedStake = utils.reduce(delegate.vaults, function(acc, _, vault)
+					return acc + vault.balance
+				end, 0),
+			})
+		end
+	end
```
dtfiedler, line 2224: don't think there's any way to avoid the O(n^2) complexity here without bloating global state in other ways. Ideally we'd look to memorize or use more efficient data structure for lookups (hash tables, maps, etc.). It may be useful to add `totalVaultedStake` attribute to gateways to avoid the additional reduction, but that would take some effort. Callers will just have to face the potential they will hit memory/CPU limts on CUs that enforce lower limits.

dtfiedler (approved):

src/gar.lua
fedellen: blue sky memory thinking over the weekend -- ideally here we'd have memoized ordered array of gateway keys and the keys would be smaller than the current addresses. so then we could paginate on gatewayKey_delegateAddress, remove sorting, and never pull down the whole gateways table only the ones we need. it is all a big lift though

PR #325, opened 2025-01-21
feat(get all vaults): add handler PE-7220

fedellen:


src/main.lua
```diff
@@ -2603,4 +2604,10 @@ addEventingHandler("allPaginatedDelegates", utils.hasMatchingTag("Action", "All-
 	Send(msg, { Target = msg.From, Action = "All-Delegates-Notice", Data = json.encode(result) })
 end)
 
+addEventingHandler("allPaginatedGatewayVaults", utils.hasMatchingTag("Action", "All-Gateway-Vaults"), function(msg)
+	local page = utils.parsePaginationTags(msg)
+	local result = gar.getPaginatedVaultsFromAllGateways(page.cursor, page.limit, page.sortBy, page.sortOrder)
+	Send(msg, { Target = msg.From, Action = "All-Delegates-Notice", Data = json.encode(result) })
```
dtfiedler, line 2610:
```suggestion
	Send(msg, { Target = msg.From, Action = "All-Gateway-Vaults-Notice", Data = json.encode(result) })
```

src/gar.lua
```diff
@@ -2233,4 +2233,37 @@ function gar.getPaginatedDelegatesFromAllGateways(cursor, limit, sortBy, sortOrd
 	)
 end
 
+--- @param cursor string|nil -- cursorId of the last item in the previous page
+--- @param limit number
+--- @param sortBy string|nil
```
arielmelendez, line 2238: Do the type docs let us narrow the valid strings down?

src/gar.lua
```diff
@@ -2233,4 +2233,37 @@ function gar.getPaginatedDelegatesFromAllGateways(cursor, limit, sortBy, sortOrd
...
+--- @param sortOrder string|nil
+--- @return PaginatedTable
+function gar.getPaginatedVaultsFromAllGateways(cursor, limit, sortBy, sortOrder)
+	--- @type DelegatesFromAllGateways[]
```
arielmelendez, line 2242: Confusing type choice since this is for gateway vaults rather than delegate vaults?

fedellen: ah good catch, some copy paste here

fedellen: dff38b7

src/gar.lua
fedellen: aaf5183

dtfiedler (approved):

PR #328, opened 2025-01-22
fix(epochs): publish the full array of prescribed observers on distribution notices

dtfiedler:


src/arns.lua
```diff
@@ -664,7 +664,7 @@ function arns.getTokenCost(intendedAction)
 		-- stub the process id as it is not required for this intent
 		local processId = "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
 		arns.assertValidBuyRecord(name, years, purchaseType, processId, false)
-		tokenCost = arns.calculateRegistrationFee(purchaseType, baseFee, years, demand.getDemandFactor())
+		tokenCost = arns.calculateRegistrationFee(purchaseType or "lease", baseFee, years, demand.getDemandFactor())
```
arielmelendez, line 667: Might make sense to hoist the `or 'lease'` up to the assignment to purchaseType to avoid potential mismatches if used elsewhere in the function.

src/epochs.lua
```diff
@@ -712,7 +727,7 @@ function epochs.distributeRewardsForEpoch(currentTimestamp)
 				)
 				if actualOperatorReward > 0 then
 					-- distribute the rewards to the gateway
-					balances.transfer(gatewayAddress, ao.id, actualOperatorReward)
+					balances.transfer(gatewayAddress, ao.id, actualOperatorReward, false) -- unsafe address should not be allowed to delegate to gateways
```
arielmelendez, line 730: This is true today, but might not be in the future. Is it important to assert that here?

src/main.lua
```diff
@@ -2610,4 +2610,21 @@ addEventingHandler("allPaginatedGatewayVaults", utils.hasMatchingTag("Action", "
 	Send(msg, { Target = msg.From, Action = "All-Gateway-Vaults-Notice", Data = json.encode(result) })
 end)
 
+-- Temporary handler for posting epoch distribution notices (make sure to remove this once previous epochs are distributed)
```
arielmelendez, line 2613: May want to add TODO here for easier discovery later

src/main.lua
```diff
@@ -2610,4 +2610,21 @@ addEventingHandler("allPaginatedGatewayVaults", utils.hasMatchingTag("Action", "
...
+addEventingHandler("postEpochDistributionNotice", utils.hasMatchingTag("Action", "Post-Epochs"), function(msg)
+	-- iterate over epochs and post epoch-distribution-notice for each epoch
+	for epochIndex in pairs(epochs.getEpochs()) do
```
arielmelendez, line 2616: Does the linter remove `, _` if you were to specify it for the unused values?

src/epochs.lua
dtfiedler: we get the linting error as the function expects this to be set

src/main.lua
dtfiedler: it's indifferent, added here for verbosity [c229f56](https://github.com/ar-io/ar-io-network-process/pull/328/commits/c229f56199a82931cd0951f2efbc643a6add8525)

src/main.lua
dtfiedler: [c229f56](https://github.com/ar-io/ar-io-network-process/pull/328/commits/c229f56199a82931cd0951f2efbc643a6add8525)

src/arns.lua
dtfiedler: ah, missed that - [c229f56](https://github.com/ar-io/ar-io-network-process/pull/328/commits/c229f56199a82931cd0951f2efbc643a6add8525)

src/epochs.lua
dtfiedler: i suppose you're correct. my thought process was that delegates are always"safe" addresses. but by introducing "allow-unsafe-addresses" anyone (including unsafe addresses) can potentially delegate, meaning rewards should be allowed to be delegated to those addresses. if this is false, it will prevent the distribution to those unsafe addresses, which is not what we want. will update.

dtfiedler: [fd61aac](https://github.com/ar-io/ar-io-network-process/pull/328/commits/fd61aacf458304d83657bf3bd80593d4ce8b1f1d)

arielmelendez (approved):

PR #330, opened 2025-01-26
chore(test): update monitor test, remove duplicate calls

dtfiedler:
These tests take ~10-15 mins to run in CI. That is too long. This leverages pLimit to concurrently execute tests, and moves some API calls up to avoid calling them multiple times across tests. There are more improvements we can make to run them concurrently (splitting into separate files), but this is sufficient for now.

These now run in 30-40 seconds on testnet.

```
ℹ tests 13
ℹ suites 9
ℹ pass 13
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 35717.852429
✨  Done in 37.58s.
```

### DIFF
--- a/tests/monitor/monitor.test.mjs
+++ b/tests/monitor/monitor.test.mjs
@@ -469,7 +469,6 @@ describe('setup', () => {
       const { items: delegates } = await io.getAllDelegates({
         limit: 10_000,
       });
-      // this overwhelms the local CU, so keep in a for loop
       await Promise.all(
         delegates.map((delegate) =>
           throttle(async () => {
--- a/tests/monitor/monitor.test.mjs
+++ b/tests/monitor/monitor.test.mjs
@@ -3,6 +3,7 @@ import { connect } from '@permaweb/aoconnect';
 import { strict as assert } from 'node:assert';
 import { describe, it, before, after } from 'node:test';
 import { DockerComposeEnvironment, Wait } from 'testcontainers';
+import pLimit from 'p-limit';
 
 // set debug level logs for to get detailed messages
 Logger.default.setLogLevel('info');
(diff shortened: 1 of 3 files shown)


tests/monitor/monitor.test.mjs
```diff
@@ -29,9 +31,8 @@ describe('setup', () => {
       projectRootPath,
       'tests/monitor/docker-compose.test.yml',
     )
-      .withBuild()
```
dtfiedler, line 32: we don't need to build everytime like we do in ar-io-node, this ends up costing 2-5 mins on every execution

tests/monitor/monitor.test.mjs
```diff
@@ -372,230 +396,187 @@ describe('setup', () => {
...
+        limit: 10_000,
       });
-      // NOTE: this is an attempt to get the delegates for a gateway, the UX is not great so consider modifying or adding handlers to support gateway specific delegation filters
-      for (const gateway of gateways) {
-        const { items: delegates } = await io.getGatewayDelegates({
-          address: gateway.gatewayAddress,
-        });
-        if (delegates.length > 0) {
-          for (const delegate of delegates) {
+      // this overwhelms the local CU, so keep in a for loop
```
arielmelendez, line 472: Was this a reminder to go back to the loop?

dtfiedler: no i will remove - i changed the test assertion given the new API

dtfiedler: [ef52d7b](https://github.com/ar-io/ar-io-network-process/pull/330/commits/ef52d7bf91559dc4aaaefe6c9ef3752681cec4ce)

PR #333, opened 2025-01-27
chore(epochs): remove old epochs

dtfiedler:
Now that we post epoch distributions we no longer need to store them in state. This should give us material memory savings, but we will validate using lua memory collector in devnet before we cleanup other functions related to pruning.

### DIFF
--- a/spec/epochs_spec.lua
+++ b/spec/epochs_spec.lua
@@ -47,7 +47,6 @@ describe("epochs", function()
 			durationMs = 100,
 			distributionDelayMs = 15,
 			rewardPercentage = 0.0025, -- 0.25%
-			pruneEpochsCount = 14,
 		}
 	end)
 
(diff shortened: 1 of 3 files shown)


dtfiedler:
I'll start with this PR - and validate the epochs are correctly pruned in devnet. Then I will remove the additional utilities around traversing old epochs

src/epochs.lua
```diff
@@ -791,11 +786,11 @@ function epochs.pruneEpochs(timestamp)
 	--- Reset the next pruning timestamp
 	NextEpochsPruneTimestamp = nil
 	local currentEpochIndex = epochs.getEpochIndexForTimestamp(timestamp)
-	local cutoffEpochIndex = currentEpochIndex - epochs.getSettings().pruneEpochsCount
+	local cutoffEpochIndex = currentEpochIndex - 1 -- keep the two most recent epochs, on distribution the last epoch will remove itself
```
dtfiedler, line 789:
```suggestion
	local cutoffEpochIndex = currentEpochIndex - 1 -- keep the two most recent epochs, on distribution the previous epoch will remove itself
```

spec/epochs_spec.lua
```diff
@@ -794,8 +797,11 @@ describe("epochs", function()
 
 			local rewardsForDelegatesThatLeft = 12500000000
 
-			-- check the epoch was updated
-			local distributions = epochs.getEpoch(epochIndex).distributions
+			-- check the epoch was properly distributed
+			if not distributedEpoch then
+				error("Distributed epoch is nil")
```
arielmelendez, line 802: Why not just assert not nil? Just wondering if there's a logical difference here.

arielmelendez (approved):

spec/epochs_spec.lua
dtfiedler: linting - it's not smart enough to recognize assert.is_not_nil passing means the object is not nil

PR #334, opened 2025-01-27
PE-7514: revoked vaults

fedellen:
this PR:

- on vaulted-transfer: when in Tags: `{name: "Revokable": value: "true"}` -- add `controller` property to the vault populated with `msg.From` (signer of the message)
- adds new new revoke-vault handler

### DIFF
--- a/src/main.lua
+++ b/src/main.lua
@@ -786,7 +786,7 @@ addEventingHandler(ActionMap.RevokeVault, utils.hasMatchingTag("Action", ActionM
 	assert(utils.isValidAddress(vaultId, true), "Invalid vault id")
 	assert(utils.isValidAddress(recipient, true), "Invalid recipient")
 
-	local vault = vaults.revokeVaultedTransfer(msg.From, recipient, vaultId, msg.Timestamp)
+	local vault = vaults.revokeVault(msg.From, recipient, vaultId, msg.Timestamp)
 
 	msg.ioEvent:addField("Vault-Id", vaultId)
 	msg.ioEvent:addField("Vault-Recipient", recipient)
--- a/src/main.lua
+++ b/src/main.lua
@@ -72,6 +72,7 @@ local ActionMap = {
 	VaultedTransfer = "Vaulted-Transfer",
 	ExtendVault = "Extend-Vault",
 	IncreaseVault = "Increase-Vault",
+	RevokeVault = "Revoke-Vault",
 	-- GATEWAY REGISTRY READ APIS
 	Gateway = "Gateway",
 	Gateways = "Gateways",
@@ -718,6 +719,7 @@ addEventingHandler(ActionMap.VaultedTransfer, utils.hasMatchingTag("Action", Act
 	local lockLengthMs = msg.Tags["Lock-Length"]
 	local msgId = msg.Id
 	local allowUnsafeAddresses = msg.Tags["Allow-Unsafe-Addresses"] or false
+	local revokable = msg.Tags.Revokable or false
 	assert(utils.isValidAddress(recipient, allowUnsafeAddresses), "Invalid recipient")
 	assert(
 		lockLengthMs and lockLengthMs > 0 and utils.isInteger(lockLengthMs),
@@ -729,14 +731,23 @@ addEventingHandler(ActionMap.VaultedTransfer, utils.hasMatchingTag("Action", Act
 	)
 	assert(recipient ~= msg.From, "Cannot transfer to self")
 
-	local vault =
-		vaults.vaultedTransfer(msg.From, recipient, quantity, lockLengthMs, msg.Timestamp, msgId, allowUnsafeAddresses)
+	local vault = vaults.vaultedTransfer(
+		msg.From,
+		recipient,
+		quantity,
+		lockLengthMs,
+		msg.Timestamp,
+		msgId,
+		allowUnsafeAddresses,
+		revokable
+	)
 
-	if vault ~= nil then
-		msg.ioEvent:addField("Vault-Id", msgId)
-		msg.ioEvent:addField("Vault-Balance", vault.balance)
-		msg.ioEvent:addField("Vault-Start-Timestamp", vault.startTimestamp)
-		msg.ioEvent:addField("Vault-End-Timestamp", vault.endTimestamp)
+	msg.ioEvent:addField("Vault-Id", msgId)
+	msg.ioEvent:addField("Vault-Balance", vault.balance)
+	msg.ioEvent:addField("Vault-Start-Timestamp", vault.startTimestamp)
+	msg.ioEvent:addField("Vault-End-Timestamp", vault.endTimestamp)
+	if revokable then
+		msg.ioEvent:addField("Vault-Controller", msg.From)
 	end
 
 	LastKnownLockedSupply = LastKnownLockedSupply + quantity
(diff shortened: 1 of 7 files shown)


src/vaults.lua
```diff
@@ -1,28 +1,30 @@
...
+--- @field controller WalletAddress | nil The controller of a revokable vault. Nil if not revokable (default: nil)
+--- @field startTimestamp Timestamp The start timestamp of the vault
+--- @field endTimestamp Timestamp The end timestamp of the vault
+
+--- @alias Vaults table<WalletAddress, table<VaultId, Vault>> -- A table of vaults indexed by owner address, then by vault id
+
+--- @type Vaults
+Vaults = Vaults or {}
 
---- @class Vaults: table<string, Vault> A table of vaults indexed by owner address
```
arielmelendez, line 17: Ah, was this improperly typed before?

src/vaults.lua
```diff
@@ -1,28 +1,30 @@
...
+NextBalanceVaultsPruneTimestamp = NextBalanceVaultsPruneTimestamp or 0
 
 --- Creates a vault
 --- @param from string The address of the owner
 --- @param qty number The quantity of tokens to vault
 --- @param lockLengthMs number The lock length in milliseconds
 --- @param currentTimestamp number The current timestamp
 --- @param vaultId string The vault id
---- @return Vault The created vault
+--- @return Vault -- The created vault
```
arielmelendez, line 27: We've used a # elsewhere before the return explanation. Not sure if `--` also parses the same way in the IDE

src/vaults.lua
```diff
@@ -53,8 +55,18 @@ end
...
+	qty,
+	lockLengthMs,
+	currentTimestamp,
+	vaultId,
+	allowUnsafeAddresses,
+	revokable
+)
 	assert(utils.isValidAddress(recipient, allowUnsafeAddresses), "Invalid recipient")
 	assert(qty > 0, "Quantity must be greater than 0")
 	assert(recipient ~= from, "Cannot transfer to self")
```
arielmelendez, line 72: Just a head check - when we call this function we should make sure that both of these fields are normalized and/or normalize here to reduce the possibility of user error.

src/vaults.lua
```diff
@@ -53,8 +55,18 @@ end
 --- @param currentTimestamp number The current timestamp
 --- @param vaultId string The vault id
 --- @param allowUnsafeAddresses boolean|nil Whether to allow unsafe addresses, since this results in funds eventually being sent to an invalid address
---- @return Vault The created vault
-function vaults.vaultedTransfer(from, recipient, qty, lockLengthMs, currentTimestamp, vaultId, allowUnsafeAddresses)
+--- @param revokable boolean|nil Whether the vault is revokable
```
arielmelendez, line 58: Should indicate what the default value is.

src/vaults.lua
```diff
@@ -53,8 +55,18 @@ end
...
+--- @return Vault -- The created vault
```
arielmelendez, line 59: Same nit about the return type comments

src/vaults.lua
```diff
@@ -74,10 +86,28 @@ function vaults.vaultedTransfer(from, recipient, qty, lockLengthMs, currentTimes
...
 	return newVault
 end
 
+--- Revokes a vaulted transfer
+---@param from WalletAddress
+---@param recipient WalletAddress
+---@param vaultId VaultId
+---@param currentTimestamp Timestamp
+---@return Vault
+function vaults.revokeVaultedTransfer(from, recipient, vaultId, currentTimestamp)
```
arielmelendez, line 100: Nit: `revokeVault` is probably a sufficient name.

src/vaults.lua
```diff
@@ -74,10 +86,28 @@ function vaults.vaultedTransfer(from, recipient, qty, lockLengthMs, currentTimes
 		balance = qty,
 		startTimestamp = currentTimestamp,
 		endTimestamp = currentTimestamp + lockLengthMs,
+		controller = revokable and from or nil,
 	})
 	return newVault
 end
 
+--- Revokes a vaulted transfer
+---@param from WalletAddress
```
arielmelendez, line 95: `from` and `recipient` are somewhat ambiguous in this use case. I don't like any of the other terms I'm thinking of better so simply adding some more context here in the descriptions might be nice.

src/vaults.lua
```diff
@@ -74,10 +86,28 @@ function vaults.vaultedTransfer(from, recipient, qty, lockLengthMs, currentTimes
...
+---@param recipient WalletAddress
+---@param vaultId VaultId
+---@param currentTimestamp Timestamp
+---@return Vault
+function vaults.revokeVaultedTransfer(from, recipient, vaultId, currentTimestamp)
+	local vault = vaults.getVault(recipient, vaultId)
+	assert(vault, "Vault not found.")
+	assert(vault.controller == from, "Only the controller can revoke the vault.")
+	assert(currentTimestamp <= vault.endTimestamp, "Vault has ended.")
```
arielmelendez, line 104: I think technically if it were `=` then the prune before this is called would have run. I also don't think revoking at the time at which the vault now belongs to the recipient is fair. Consider `<` only.

src/vaults.lua
```diff
@@ -74,10 +86,28 @@ function vaults.vaultedTransfer(from, recipient, qty, lockLengthMs, currentTimes
...
+
+	balances.increaseBalance(recipient, vault.balance)
```
arielmelendez, line 106: The balance is supposed to go to the `from`! This was my concern about the ambiguity in the argument names.

tests/vaults.test.mjs
```diff
@@ -338,10 +339,81 @@ describe('Vaults', async () => {
...
+          (tag) => tag.name === 'Action' && tag.value === 'Revoke-Vault-Notice',
+        ),
+      );
+      assert.ok(controllerMessageAfterRevoke);
+      const controllerVaultDataAfterRevoke = JSON.parse(
+        controllerMessageAfterRevoke.Data,
+      );
+      assert.deepEqual(controllerVaultDataAfterRevoke, expectedVaultData);
+
+      endingMemory = result.Memory;
```
arielmelendez, line 412: This test needs to check the balance of the revoker for increase. If here it would have caught the logical error noted above.

src/main.lua
```diff
@@ -718,6 +719,7 @@ addEventingHandler(ActionMap.VaultedTransfer, utils.hasMatchingTag("Action", Act
 	local lockLengthMs = msg.Tags["Lock-Length"]
 	local msgId = msg.Id
 	local allowUnsafeAddresses = msg.Tags["Allow-Unsafe-Addresses"] or false
+	local revokable = msg.Tags.Revokable or false
```
arielmelendez, line 722: Be careful with boolean tag parameters. Do they get passed through to here as boolean or as strings? If strings then you'll need a comparison to "true".

src/main.lua
```diff
@@ -729,14 +731,23 @@ addEventingHandler(ActionMap.VaultedTransfer, utils.hasMatchingTag("Action", Act
 	)
 	assert(recipient ~= msg.From, "Cannot transfer to self")
 
-	local vault =
-		vaults.vaultedTransfer(msg.From, recipient, quantity, lockLengthMs, msg.Timestamp, msgId, allowUnsafeAddresses)
+	local vault = vaults.vaultedTransfer(
+		msg.From,
```
arielmelendez, line 735: Make sure from and recipient are passed through normalized. My quick reading of the code is the choice of using msg.From and msg.Tags.Recipient should be normalized by the sanitization handler.

tests/vaults.test.mjs
arielmelendez: you could/should also check for a credit-notice sent to sender.

arielmelendez (changes requested):
Some logical issues to sort out.

src/vaults.lua
dtfiedler: +1

src/vaults.lua
fedellen: `--` and `#` result in the same IDE output for me: 
image

src/vaults.lua
fedellen: assertAndSanitize util formats the addresses before getting to the handler

src/main.lua
fedellen: yes, they also get formatted in assertAndSanitize

src/vaults.lua
fedellen: yikes, great catch

fixed unit coverage here, unluckily I used an address that already has 100 balance in the test env: [1835ad6](https://github.com/ar-io/ar-io-network-process/pull/334/commits/1835ad63b2aa4dead92ec60fc1850a6fcc55b4f9)

actual fix here for red/green in CI: [c8ed870](https://github.com/ar-io/ar-io-network-process/pull/334/commits/c8ed870db47188043ce463ff144ff6b97a18401a)

image

src/vaults.lua
fedellen: updated in this commit: [c8ed870](https://github.com/ar-io/ar-io-network-process/pull/334/commits/c8ed870db47188043ce463ff144ff6b97a18401a)

src/vaults.lua
fedellen: [2ca8dce](https://github.com/ar-io/ar-io-network-process/pull/334/commits/2ca8dcea8aef1b18d91e3aebbc920f5645474085)

src/vaults.lua
fedellen: [2ca8dce](https://github.com/ar-io/ar-io-network-process/pull/334/commits/2ca8dcea8aef1b18d91e3aebbc920f5645474085)

src/vaults.lua
fedellen: they result in the same output. I'd update them all across the codebase but `--` is a hard grep to search and replace on

tests/vaults.test.mjs
fedellen: 881b3c6

src/vaults.lua
fedellen: yes, there were a couple type issues in the IDE that were resolved from this

PR #338, opened 2025-01-28
fix(epochs): ensure previous epoch has been distributed before creati…

dtfiedler:
…ng new one

A bit more cleanup around handling old epochs. A slight modification for the logic around creating a new epoch, as well is returning `nil` for epochs that do not exist as opposed to an empty object.

### DIFF
--- a/spec/epochs_spec.lua
+++ b/spec/epochs_spec.lua
@@ -458,131 +458,128 @@ describe("epochs", function()
 	end)
 
 	describe("createEpoch", function()
-		it(
-			"should create a new epoch for the given timestamp if there is no previous epoch that has not been distributed",
-			function()
-				_G.Epochs = {}
-				_G.GatewayRegistry = {
-					["test-this-is-valid-arweave-wallet-address-1"] = {
-						operatorStake = gar.getSettings().operators.minStake,
-						totalDelegatedStake = 0,
-						vaults = {},
-						delegates = {},
-						startTimestamp = 0,
-						stats = {
-							prescribedEpochCount = 0,
-							observedEpochCount = 0,
-							totalEpochCount = 0,
-							passedEpochCount = 0,
-							failedEpochCount = 0,
-							failedConsecutiveEpochs = 0,
-							passedConsecutiveEpochs = 0,
-						},
-						settings = testSettings,
-						status = "joined",
-						observerAddress = "test-this-is-valid-arweave-wallet-address-1",
-						weights = {
-							stakeWeight = 0,
-							tenureWeight = 0,
-							gatewayRewardRatioWeight = 0,
-							observerRewardRatioWeight = 0,
-							compositeWeight = 0,
-							normalizedCompositeWeight = 0,
-						},
+		it("should create a new epoch for the given timestamp if all previous epochs have been distributed", function()
+			_G.Epochs = {}
+			_G.GatewayRegistry = {
+				["test-this-is-valid-arweave-wallet-address-1"] = {
+					operatorStake = gar.getSettings().operators.minStake,
+					totalDelegatedStake = 0,
+					vaults = {},
+					delegates = {},
+					startTimestamp = 0,
+					stats = {
+						prescribedEpochCount = 0,
+						observedEpochCount = 0,
+						totalEpochCount = 0,
+						passedEpochCount = 0,
+						failedEpochCount = 0,
+						failedConsecutiveEpochs = 0,
+						passedConsecutiveEpochs = 0,
 					},
-					-- not eligible for rewards as it is leaving, so it should not be included in the eligible gateways count
-					["test-this-is-valid-arweave-wallet-address-2"] = {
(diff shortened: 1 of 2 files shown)


src/epochs.lua
```diff
@@ -474,6 +461,7 @@ function epochs.saveObservations(observerAddress, reportTxId, failedGatewayAddre
 	assert(observingGateway, "The associated gateway not found in the registry.")
 
 	local epoch = epochs.getEpoch(epochIndex)
+	assert(epoch, "Unable to save observation. Epoch not found for index: " .. epochIndex)
```
dtfiedler, line 464: modified as epoch could be `nil` instead of `{}`

src/epochs.lua
```diff
@@ -598,25 +586,26 @@ end
 function epochs.distributeRewardsForEpoch(currentTimestamp)
 	local epochIndex = epochs.getEpochIndexForTimestamp(currentTimestamp - epochs.getSettings().durationMs) -- go back to previous epoch
 	local epoch = epochs.getEpoch(epochIndex)
-	if not next(epoch) then
+	if not epoch then
 		-- silently return
-		print("Not distributing rewards for last epoch. New epoch has not been created yet.")
+		print("Unable to distribute rewards for epoch. Epoch not found: " .. epochIndex)
```
dtfiedler, line 591: considering making this an error instead of a silent return

src/epochs.lua
```diff
@@ -598,25 +586,26 @@ end
...
 	end
 
 	--- The epoch was already distributed
 	--- @cast epoch DistributedEpoch
 	if epoch.distributions.distributedTimestamp then
-		print("Rewards already distributed for epoch: " .. epochIndex)
-		return -- silently return
+		print("Rewards already distributed for epoch. Epoch will be removed from the epoch registry: " .. epochIndex)
+		Epochs[epochIndex] = nil
+		return epoch
```
dtfiedler, line 600: this mean the cleanup didn't happen, so do it here and remove the distributed epoch.

spec/epochs_spec.lua
```diff
@@ -459,17 +459,9 @@ describe("epochs", function()
 
 	describe("createEpoch", function()
 		it(
-			"should create a new epoch for the given timestamp once distributions for the last epoch have occurred",
+			"should create a new epoch for the given timestamp if there is no previous epoch that has not been distributed",
```
arielmelendez, line 462: oof... double negatives....

arielmelendez: Maybe: `... if all previous epics have been distributed`?

src/epochs.lua
```diff
@@ -370,26 +364,19 @@ function epochs.createEpoch(timestamp, blockHeight, hashchain)
 	assert(type(hashchain) == "string", "Hashchain must be a string")
 
 	local epochIndex = epochs.getEpochIndexForTimestamp(timestamp)
-	if next(epochs.getEpoch(epochIndex)) then
+	if epochs.getEpoch(epochIndex) then
 		-- silently return
-		print("Epoch already exists for index: " .. epochIndex)
-		return
+		print("Epoch already exists for index. Returning existing epoch: " .. epochIndex)
+		return nil
```
arielmelendez, line 370: The comment above says "returning existing" but we're returning nil. Is that intended?

src/epochs.lua
arielmelendez: Why the hesitation?

src/epochs.lua
arielmelendez: Does returning the epoch rather than nil have any unintended downstream repercussions?

arielmelendez (approved):

src/epochs.lua
dtfiedler: breaking tick being the primary concern. i'd like to have this run in devnet and validate before introducing error states

src/epochs.lua
dtfiedler: it means a send-distribution-notice is sent, which is what we want if the epoch was not previously cleared

src/epochs.lua
dtfiedler: modified here with a note why not to return the epoch - [c8472ba](https://github.com/ar-io/ar-io-network-process/pull/338/commits/c8472ba416686d262eda4e32d5819957931464d9)

spec/epochs_spec.lua
dtfiedler: [c8472ba](https://github.com/ar-io/ar-io-network-process/pull/338/commits/c8472ba416686d262eda4e32d5819957931464d9)

src/epochs.lua
dtfiedler: added a TODO to make this an error state - [c8472ba](https://github.com/ar-io/ar-io-network-process/pull/338/commits/c8472ba416686d262eda4e32d5819957931464d9)

arielmelendez (approved):

PR #341, opened 2025-01-29
chore(PE-7040): replace gatewayRewardRatioWeight and observerRewardRatioWeight with gatewayPerformanceRatio and observerPerformanceRatio

dtfiedler:
Swap out the attribute name. for the handlers we'll support the old ones until the SDK and network portal are updated.

https://github.com/ar-io/ar-io-sdk/pull/367


dtfiedler:
I've backfilled this on devnet and everything looks good. the backwards compatibility can be removed once the SDK is pushed through the network portal.

fedellen (approved):

PR #342, opened 2025-01-29
fix(epochs): schedule next epochs pruning

dtfiedler:


dtfiedler:
i don't think this is necessary - epochs clean themselves up and the current state of things will be resolved on the next distribution

PR #343, opened 2025-01-29
fix(epochs): do not return previously distributed epoch if already di…

dtfiedler:
…stributed

This should have been done in previous PR. There are side effects when distributing epochs (event data, and notices) so we DO NOT want to return them if they were already distributed. We do want to remove them from the registry if they exist and have been distributed, however.

### DIFF
--- a/src/epochs.lua
+++ b/src/epochs.lua
@@ -587,17 +587,15 @@ function epochs.distributeRewardsForEpoch(currentTimestamp)
 	local epochIndex = epochs.getEpochIndexForTimestamp(currentTimestamp - epochs.getSettings().durationMs) -- go back to previous epoch
 	local epoch = epochs.getEpoch(epochIndex)
 	if not epoch then
-		-- TODO: consider throwing an error here instead of silently returning, as this is a critical error and should be fixed
-		print("Unable to distribute rewards for epoch. Epoch not found: " .. epochIndex)
-		return
+		error("Unable to distribute rewards for epoch. Epoch does not exist in epoch registry: " .. epochIndex)
 	end
 
-	--- The epoch was already distributed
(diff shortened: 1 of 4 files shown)


src/epochs.lua
```diff
@@ -374,12 +380,12 @@ function epochs.createEpoch(timestamp, blockHeight, hashchain)
 	local prevEpoch = epochs.getEpoch(prevEpochIndex)
 	-- if there is a previous epoch and it has not been distributed, we cannot create a new epoch. once the epoch has been distributed, it will be removed from the epoch registry
 	if prevEpoch and not prevEpoch.distributions.distributedTimestamp then
-		-- silently return
+		-- TODO: consider throwing an error here instead of silently returning
```
dtfiedler, line 383: these will require some additional test changes, so will address separately

src/epochs.lua
```diff
@@ -116,7 +116,8 @@ end
 --- @param epochIndex number The epoch index
 --- @return table<WalletAddress, WalletAddress> # The prescribed observers for the epoch
 function epochs.getPrescribedObserversForEpoch(epochIndex)
-	return epochs.getEpoch(epochIndex).prescribedObservers or {}
+	local epoch = epochs.getEpoch(epochIndex)
+	return epoch and epoch.prescribedObservers or {}
```
arielmelendez, line 120: I'm assuming that returning empty table rather than nil for this and all the below is preferable for flow control and logic management?

arielmelendez (approved):

src/epochs.lua
dtfiedler: correct - right now, it simplifies logic when traversing these tables. the impact of changing them to `nil` would be larger change, but not unreasonable.

PR #347, opened 2025-02-01
chore(PE-7555): fix arns pruning bug, standardize logic for pruning

dtfiedler:
This bug delays the pruning of records that should be removed after their grace period ends. 

TLDR:
- we correctly remove records that are expired
- we correctly update the minimum `NextRecordsPruneTimestamp` for those in their grace period
- we use `minEndTimestamp` to track the lowest endTimestamp of all records that are NOT in their grace period or expired
- we INCORRECTLY `nil` out `NextRecordsPruneTimestamp` and then set it to the `minNextEndTimestamp` variable if it was set, **overwriting any minimum that should have been set by records in their grace period**

Here is the `pruneRecords` function with additional comments (IN CAPS) on where/why the bug occurs

```lua
--- Prunes records that have expired
--- @param lastGracePeriodEntryEndTimestamp number The end timestamp of the last known record to have entered its grace period
--- @return table<string, Record> prunedRecords - the pruned records
--- @return table<string, Record> recordsInGracePeriod - the records that have entered their grace period
function arns.pruneRecords(currentTimestamp, lastGracePeriodEntryEndTimestamp)
	lastGracePeriodEntryEndTimestamp = lastGracePeriodEntryEndTimestamp or 0
	local prunedRecords = {}
	local newGracePeriodRecords = {}
       -- WHEN THE NextRecordsPruneTimestamp IS SET INCORRECTLY BELOW, WE DELAY THE PRUNING OF RECORDS THAT SHOULD HAVE BEEN REMOVED EARLIER
	if not NextRecordsPruneTimestamp or NextRecordsPruneTimestamp > currentTimestamp then
		return prunedRecords, newGracePeriodRecords
	end

	--- @type Timestamp|nil
	local minNextEndTimestamp -- THIS LEADS TO THE BUG

        for name, record in pairs(arns.getRecords()) do
		if arns.recordExpired(record, currentTimestamp) then
			prunedRecords[name] = record
			NameRegistry.records[name] = nil
		elseif arns.recordInGracePeriod(record, currentTimestamp) then
			if record.endTimestamp > lastGracePeriodEntryEndTimestamp then
				newGracePeriodRecords[name] = record
			end
                        --- WE UPDATE TO THE GRACE PERIOD CORRECTLY HERE, AND IF THE RECORD IS LAST ONE TRAVERSED, WE'RE ALL GOOD!
			arns.scheduleNextRecordsPrune(record.endTimestamp + constants.gracePeriodMs)
		elseif record.endTimestamp then
                        --- EVERY OTHER RECORD THAT IS NOT EXPIRED/IN GRACE PERIOD UPDATES THIS VALUE
                        --- IT WILL USUALLY BE GREATER THAN THE LOWEST DETERMINED BY THE GRACE PERIOD RECORDS ABOVE!
			minNextEndTimestamp = math.min(minNextEndTimestamp or record.endTimestamp, record.endTimestamp)
		end
	end

        -- WE NIL OUT THE GLOBAL & minNextEndTimestamp IS THE LOWEST ACTIVE RECORD END TIMESTAMP
        -- THIS CAUSES US TO SKIP OVER THE MINIMUM THAT SHOULD HAVE BEEN SET BY RECORDS IN THEIR GRACE PERIOD!
	NextRecordsPruneTimestamp = nil
	if minNextEndTimestamp then
		arns.scheduleNextRecordsPrune(minNextEndTimestamp)
	end
	return prunedRecords, newGracePeriodRecords
end
```

This pattern is true for all prunes, so they have been updated in this PR to avoid using the minEndTimestamp, nil out the global first, then run the iteration to identify the minimum prune timestamp.

### DIFF
--- a/spec/arns_spec.lua
+++ b/spec/arns_spec.lua
@@ -1026,13 +1026,60 @@ describe("arns", function()
 	end)
 
 	describe("pruneRecords", function()
-		it("should prune records older than the grace period", function()
-			local currentTimestamp = 2000000000
+		it(
+			"should prune records older than the grace period and update the `NextRecordsPruneTimestamp` to the minimum record end timestamp plus grace period",
+			function()
+				local currentTimestamp = 2000000000
 
-			_G.NameRegistry = {
-				returned = {},
-				reserved = {},
-				records = {
+				_G.NameRegistry = {
+					returned = {},
+					reserved = {},
+					records = {
+						["active-record"] = {
+							endTimestamp = 2001000000, -- far in the future
+							processId = "active-process-id",
+							purchasePrice = 600000000,
+							startTimestamp = 0,
+							type = "lease",
+							undernameLimit = 10,
+						},
+						["active-record-2"] = {
+							endTimestamp = 10005000000, -- far in the future
+							processId = "active-process-id-2",
+							purchasePrice = 600000000,
+							startTimestamp = 0,
+							type = "lease",
+							undernameLimit = 10,
+						},
+						["expired-record"] = {
+							endTimestamp = 790399999, -- expired and past the grace period
+							processId = "expired-process-id",
+							purchasePrice = 400000000,
+							startTimestamp = 0,
+							type = "lease",
+							undernameLimit = 5,
+						},
+						["grace-period-record"] = {
+							endTimestamp = currentTimestamp - constants.gracePeriodMs + 10, -- expired, but within grace period
+							processId = "grace-process-id",
+							purchasePrice = 500000000,
+							startTimestamp = 0,
+							type = "lease",
+							undernameLimit = 8,
+						},
+						["permabuy-record"] = {
+							endTimestamp = nil,
+							processId = "permabuy-process-id",
+							purchasePrice = 600000000,
+							startTimestamp = 0,
+							type = "permabuy",
+							undernameLimit = 10,
+						},
+					},
+				}
+				local prunedRecords, newGracePeriodRecords =
+					arns.pruneRecords(currentTimestamp, _G.LastGracePeriodEntryEndTimestamp)
+				assert.are.same({
 					["active-record"] = {
 						endTimestamp = 2001000000, -- far in the future
 						processId = "active-process-id",
@@ -1041,13 +1088,13 @@ describe("arns", function()
 						type = "lease",
 						undernameLimit = 10,
 					},
-					["expired-record"] = {
-						endTimestamp = 790399999, -- expired and past the grace period
-						processId = "expired-process-id",
-						purchasePrice = 400000000,
+					["active-record-2"] = {
+						endTimestamp = 10005000000, -- very far in the future
+						processId = "active-process-id-2",
+						purchasePrice = 600000000,
 						startTimestamp = 0,
 						type = "lease",
-						undernameLimit = 5,
+						undernameLimit = 10,
 					},
 					["grace-period-record"] = {
 						endTimestamp = currentTimestamp - constants.gracePeriodMs + 10, -- expired, but within grace period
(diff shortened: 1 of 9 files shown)


src/arns.lua
```diff
@@ -977,32 +976,27 @@ function arns.pruneRecords(currentTimestamp, lastGracePeriodEntryEndTimestamp)
...
-			minNextEndTimestamp = math.min(minNextEndTimestamp or record.endTimestamp, record.endTimestamp)
+			arns.scheduleNextRecordsPrune(record.endTimestamp)
 		end
 	end
-
-	-- Reset the next pruning timestamp now that pruning has completed
-	NextRecordsPruneTimestamp = nil
-	if minNextEndTimestamp then
-		arns.scheduleNextRecordsPrune(minNextEndTimestamp)
-	end
```
dtfiedler, line 1005: this is the cause of the bug. `minNextEndTimestamp` is only set/updated for records that are NOT NOR IN THEIR GRACE PERIOD (if statement above). Because `NextRecordsPruneTimestamp` is set to `nil` above, this value will ALWAYS be updated to `minNextEndTimestamp`  when calling `arns.scheduleNextRecordsPrune`. Ultimately IGNORING records the minimum that SHOULD be set by records entering their grace period.

spec/arns_spec.lua
```diff
@@ -1026,13 +1026,60 @@ describe("arns", function()
...
 	describe("pruneRecords", function()
-		it("should prune records older than the grace period", function()
-			local currentTimestamp = 2000000000
-
-			_G.NameRegistry = {
-				returned = {},
-				reserved = {},
-				records = {
+		it(
+			"should prune records older than the grace period and update the `NextRecordsPruneTimestamp` to the minimum record end timestamp plus grace period",
```
fedellen, line 1030: 👏

tests/helpers.mjs
```diff
@@ -865,6 +865,53 @@ export const tick = async ({
...
 
+export const getInfo = async ({ memory, timestamp }) => {
+  const nameResult = await handle({
+    options: {
+      Tags: [{ name: 'Action', value: 'Info' }],
+      Timestamp: timestamp,
+    },
+    memory,
+  });
+  console.log(nameResult);
```
fedellen, line 876:
```suggestion
```

fedellen (approved):

src/arns.lua
```diff
@@ -977,32 +976,27 @@ function arns.pruneRecords(currentTimestamp, lastGracePeriodEntryEndTimestamp)
...
+	for name, record in pairs(arns.getRecordsUnsafe()) do
 		if arns.recordExpired(record, currentTimestamp) then
-			prunedRecords[name] = record
-			NameRegistry.records[name] = nil
+			print("Pruning record " .. name .. " because it has expired")
+			prunedRecords[name] = arns.removeRecord(name)
 		elseif arns.recordInGracePeriod(record, currentTimestamp) then
 			if record.endTimestamp > lastGracePeriodEntryEndTimestamp then
+				print(
+					"Adding record " .. name .. " to new grace period records because it has entered it's grace period"
```
arielmelendez, line 990: nit: `its`

dtfiedler: [c7c630d](https://github.com/ar-io/ar-io-network-process/pull/347/commits/c7c630dae9d39cd595cbf0160da0108d1fc6a73f)

src/constants.lua
```diff
@@ -53,9 +53,9 @@ constants.ARNS_NAME_DOES_NOT_EXIST_MESSAGE = "Name not found in the ArNS Registr
 constants.UNDERNAME_LEASE_FEE_PERCENTAGE = 0.001
 constants.UNDERNAME_PERMABUY_FEE_PERCENTAGE = 0.005
 constants.PRIMARY_NAME_REQUEST_COST = constants.ARIOToMARIO(10) -- 10 ARIO
-constants.gracePeriodMs = constants.defaultEpochDurationMs * 14 -- 14 epochs
+constants.gracePeriodMs = constants.twoWeeksMs
```
arielmelendez, line 56: Nit that the previous representation dealt with epic counts directly whereas this value coincidentally works out to the correct number of epochs.

src/epochs.lua
```diff
@@ -795,15 +796,17 @@ function epochs.pruneEpochs(timestamp)
 			Epochs[nextEpochIndex] = nil
 		else
 			local _, endTimestamp = epochs.getEpochTimestampsForIndex(nextEpochIndex)
-			if endTimestamp >= timestamp then
```
arielmelendez, line 798: This is a logical change, correct?

src/constants.lua
dtfiedler: grace period has no affiliation with epochs

src/gar.lua
```diff
@@ -1076,22 +1077,14 @@ function gar.pruneGateways(currentTimestamp, msgId)
 							table.insert(result.prunedGateways, gatewayAddress)
 						else
 							-- find the next prune timestamp
-							minNextEndTimestamp =
-								--- @diagnostic disable-next-line: param-type-mismatch
-								math.min(minNextEndTimestamp or gateway.endTimestamp, gateway.endTimestamp)
+							gar.scheduleNextGatewaysPruning(minNextEndTimestamp)
```
arielmelendez, line 1080: This local `minNextEndTimestamp ` was removed up top, right? And isn't the thing that needs to be considered for scheduling gateway.endTimestamp?

dtfiedler: good catch, this needs to be updated wherever we were using `minNextEndTimestamp`

dtfiedler: [f4d8ec2](https://github.com/ar-io/ar-io-network-process/pull/347/commits/f4d8ec2dbfe972e41d77cbc61c8072dae7302ebd)

src/epochs.lua
dtfiedler: reverted here - [23c1d55](https://github.com/ar-io/ar-io-network-process/pull/347/commits/23c1d553c2e474a6e76a6d4b1eed151ab33ee424)

i will write some more integration and unit tests validating these pruning behaviors in a separate PR before removing any logic

fedellen (approved):

arielmelendez (approved):

PR #352, opened 2025-02-04
fix(arns): fix arns stat issue

dtfiedler:
These stats are computed at the beginning of an epoch, and just used for historical tracking. Every epoch has a start and timestamp - and these stats are computed at the beginning on an epoch.

- Active names: at the beginning of an epoch, these names are ACTIVE
- Reserved names: at the beginnog of an epoch, these names are RESERVED
- Returned names: at the beginning of an epoch these names are RETURNED
- Grace period names: at the beginning of an epoch, these names are in their GRACE PERIOD

### DIFF
--- a/src/arns.lua
+++ b/src/arns.lua
@@ -387,12 +387,12 @@ local function getTotalArNSNamesBetweenTimestampsUnsafe(startTimestamp, endTimes
 		if record.type == "permabuy" then
 			totalActiveNames = totalActiveNames + 1
 		elseif record.type == "lease" and record.endTimestamp then
-			-- record is active between the start and end timestamps
+			-- record is active for the entire period between the two timestamps
 			if record.startTimestamp <= startTimestamp and record.endTimestamp >= endTimestamp then
 				totalActiveNames = totalActiveNames + 1
 			elseif
-				-- record is in the grace period between the start and end timestamps
-				record.startTimestamp <= startTimestamp
+				-- record is in the grace period for the entire period between the two timestamps
+				record.endTimestamp >= startTimestamp
 				and record.endTimestamp + constants.gracePeriodMs >= endTimestamp
 			then
 				totalGracePeriodNames = totalGracePeriodNames + 1
--- a/src/arns.lua
+++ b/src/arns.lua
@@ -387,9 +387,11 @@ local function getTotalArNSNamesBetweenTimestampsUnsafe(startTimestamp, endTimes
 		if record.type == "permabuy" then
 			totalActiveNames = totalActiveNames + 1
 		elseif record.type == "lease" and record.endTimestamp then
+			-- record is active between the start and end timestamps
 			if record.startTimestamp <= startTimestamp and record.endTimestamp >= endTimestamp then
 				totalActiveNames = totalActiveNames + 1
 			elseif
+				-- record is in the grace period between the start and end timestamps
 				record.startTimestamp <= startTimestamp
 				and record.endTimestamp + constants.gracePeriodMs >= endTimestamp
 			then
@@ -408,7 +410,8 @@ local function getTotalReservedNamesBetweenTimestampsUnsafe(_, endTimestamp)
 	local reservedNames = arns.getReservedNamesUnsafe()
 	local totalReservedNames = 0
 	for _, reservedName in pairs(reservedNames) do
-		if not reservedName.endTimestamp or endTimestamp <= reservedName.endTimestamp then
+		-- reserved name is active between the start and end timestamps
+		if not reservedName.endTimestamp or reservedName.endTimestamp >= endTimestamp then
 			totalReservedNames = totalReservedNames + 1
 		end
 	end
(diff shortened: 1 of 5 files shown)


src/arns.lua
```diff
@@ -415,10 +421,13 @@ end
...
 	local returnedNames = arns.getReturnedNamesUnsafe()
 	local totalReturnedNames = 0
-	local returnedNameEndTimestamp = startTimestamp + constants.returnedNamePeriod
 
 	for _, returnedName in pairs(returnedNames) do
-		if startTimestamp >= returnedName.startTimestamp and endTimestamp <= returnedNameEndTimestamp then
+		-- returned name is active between the start and end timestamps
+		if
+			returnedName.startTimestamp <= startTimestamp
+			and returnedName.startTimestamp + constants.returnedNamePeriod >= endTimestamp
```
dtfiedler, line 429: no change in logic here - just standardizing them, so it reads as : "X started BEFORE startTimestamp, and ends AFTER endTimestamp)"

fedellen (approved):

src/arns.lua
```diff
@@ -387,9 +387,14 @@ local function getTotalArNSNamesBetweenTimestampsUnsafe(startTimestamp, endTimes
 		if record.type == "permabuy" then
 			totalActiveNames = totalActiveNames + 1
 		elseif record.type == "lease" and record.endTimestamp then
+			-- record is active between the start and end timestamps
 			if record.startTimestamp <= startTimestamp and record.endTimestamp >= endTimestamp then
```
arielmelendez, line 391: Why is the start timestamp relevant here? It's not clear what this function's criteria are for names being between the timestamp. I would think that it's active if:
- record.endTimestamp is greater than startTimestamp of interest (e.g. it didn't expire before the startTimestamp of interest)
- record.endTimestamp is also less than or equal to endTimestamp of interest (still not expired)

record.startTimestamps seem kind of irrelevant depending on what's intended here.

src/arns.lua
```diff
@@ -387,9 +387,14 @@ local function getTotalArNSNamesBetweenTimestampsUnsafe(startTimestamp, endTimes
...
 				totalActiveNames = totalActiveNames + 1
-			elseif record.endTimestamp + constants.gracePeriodMs >= endTimestamp then
+			elseif
+				-- record is in the grace period between the start and end timestamps
```
arielmelendez, line 394: Grace period has to do with the CURRENT timestamp. I don't know if you can tell that something is in grace period given two arbitrary timestamps in the way that this function seems to.

src/arns.lua
dtfiedler: if you wanted to slice of records that were active in a window - you have to account for when they started. this eliminates records that start after the provided current timestamp as they do not exist for the entirety of the boundaries

src/arns.lua
dtfiedler: there is an issue with the current implementation e can only know of the record.endTimestamp > startTimestamp, so this needs to be fixed

dtfiedler: [12d396b](https://github.com/ar-io/ar-io-network-process/pull/352/commits/12d396b35436049058af2b96a18d4defac5ac0ef)

src/arns.lua
dtfiedler: i updated this to drop the boundary constraint an only concern itself with epoch start timestamp. this reduces some of the logic you asked about [560dda9](https://github.com/ar-io/ar-io-network-process/pull/352/commits/560dda98224d770dfeaacec77ea1fc9ebedb5395)

src/arns.lua
```diff
@@ -387,9 +390,9 @@ local function getTotalArNSNamesBetweenTimestampsUnsafe(startTimestamp, endTimes
 		if record.type == "permabuy" then
 			totalActiveNames = totalActiveNames + 1
 		elseif record.type == "lease" and record.endTimestamp then
-			if record.startTimestamp <= startTimestamp and record.endTimestamp >= endTimestamp then
+			if arns.recordIsActive(record, timestamp) then
 				totalActiveNames = totalActiveNames + 1
-			elseif record.endTimestamp + constants.gracePeriodMs >= endTimestamp then
+			elseif arns.recordInGracePeriod(record, timestamp) then
```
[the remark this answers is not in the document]
dtfiedler: leaning in to our existing functions instead of worrying about boundaries of an epoch ✅

fedellen (dismissed):

src/arns.lua
dtfiedler: tweaked implementation in [560dda9](https://github.com/ar-io/ar-io-network-process/pull/352/commits/560dda98224d770dfeaacec77ea1fc9ebedb5395)

src/arns.lua
```diff
@@ -925,7 +940,11 @@ function arns.recordIsActive(record, currentTimestamp)
 		return true
 	end
 
-	return record.endTimestamp and record.endTimestamp >= currentTimestamp
+	-- record starts before the current timestamp and ends after the current timestamp
+	return record.startTimestamp
+		and record.startTimestamp <= currentTimestamp
```
arielmelendez, line 945: Nit: maybe update currentTimestamp to just timestamp or atTimestamp since this is now a general purpose function that can look into the past.

arielmelendez (dismissed):

src/arns.lua
dtfiedler: [dfad08c](https://github.com/ar-io/ar-io-network-process/pull/352/commits/dfad08cb6ad883250bfd28115cd0220df81748de)

fedellen (approved):
thanks for clean ups. sorry for the churn here!

fedellen (approved):

PR #353, opened 2025-02-05
feat(epochs): remove distribution delay, start and prescribe new epochs immediately

dtfiedler:
**Before:**
```mermaid
gantt
    title Epoch Gantt Chart

    %% Epoch 1
    section Epoch 1
    Distribution Delay (account for forking)                      :a1, 2025-02-06 00:00, 3h
    Epoch Starts (back dated to start timestamp)             :a2, after a1, 2m
    Epoch Prescribed               :a3, after a2, 2m
    Observation Period            :a4, after a3, 10h
    Epoch ends                         :a5, after a4, 2m
    Distribution Delay (account for forking)              :a6, after a5, 3h
    Epoch Distributed                           :a7, after a6, 2m
    Epoch pruned                     :a8, after a7, 2m

    %% Epoch 2
    section Epoch 2
    Epoch Starts   (back dated to start timestamp)                       :b2, after a8, 2m
    Epoch Prescribed               :b3, after b2, 2m
    Observation Period               :b4, after b3, 10h
    Epoch ends                       :b5, after b4, 2m
    Distribution Delay               :b6, after b5, 3h
    Epoch Distributed              :b7, after b6, 2m
    Epoch pruned                     :b8, after b7, 2m

```

**AFTER**
```mermaid
gantt
    title Epoch Gantt Chart

    %% Epoch 1
    section Epoch 1
    Epoch Starts                       :a2,  2025-02-06 00:00, 2m
    Epoch Prescribed               :a3, after a2, 2m
    Observation Period            :a4, after a3, 10h
    Epoch ends                         :a5, after a4, 2m
    Epoch Distributed                           :a7, after a5, 2m
    Epoch pruned                     :a8, after a7, 2m

    %% Epoch 2
    section Epoch 2
    Epoch Starts                     :b2, after a7, 2m
    Epoch Prescribed               :b3, after b2, 2m
    Observation Period               :b4, after b3, 10h
    Epoch ends                       :b5, after b4, 2m
    Epoch Distributed              :b7, after b5, 2m
    Epoch pruned                     :b8, after b7, 2m

```

Current state:
- **Created**: the epoch start timestamp is hit and the initial epoch with it's respective metadata (start timestamp, end timestamp, index, block height, etc.) are all created and stored in state. 
- **Prescribed**: we wait ~40 minutes (aka the `distributionDelay`) before prescribing the current epoch. This includes selecting the observers and arns names for observation. Additionally, all the eligible rewards are established for the epoch. We do this here (instead of when we create the epoch) we need to reconcile and distribute rewards before we can allocate new ones, and observer weights are impacted by stake weights/auto-staking that change after distribution 
- **Observation**: prescribed observers submit observations throughout the epoch, all observations must be submitted by the epochs end timestamp
- **Distributed**: once the distribution timestamp of the previous epoch is reached, it is distributed. This includes allocating rewards to gateways and their delegates, and updating gateway stats based on their performance throughout the epoch 

Changes in this PR:
- removes the `distributionDelay` and distributes epochs at first tick after their end timestamp
- new epochs get created and prescribed immediately (`Epoch-Created-Notice` message is sent)


src/epochs.lua
```diff
@@ -366,79 +367,119 @@ function epochs.getEpochIndexForTimestamp(timestamp)
...
-		for _, weightedGateway in ipairs(weightedGateways) do
-			gar.updateGatewayWeights(weightedGateway)
-		end
-	end
 
 	-- Force schedule a pruning JIC
 	NextEpochsPruneTimestamp = NextEpochsPruneTimestamp or 0
 
+	-- update the weights for all active gateways
+	gar.updateAllActiveGatewayWeights(timestamp)
```
dtfiedler, line 481: consolidates this logic into a function

tests/epochs.test.mjs
dtfiedler, line 1: these tests now cover genesis epoch -> second epoch with distributions. some of the tests overlap with `tick` but the more coverage here the better

tests/epochs.test.mjs
```diff
@@ -8,148 +8,337 @@ import {
...
+            totalReservedNames: 0,
+            totalReturnedNames: 0,
+          },
+          prescribedObservers: [],
+          prescribedNames: [],
+          observations: {
+            failureSummaries: [], // TODO: ideally this is encoded as an empty object, consider adding a helper function to encode empty objects
+            reports: [], // TODO: ideally this is encoded as an empty object, consider adding a helper function to encode empty objects
+          },
+          distributions: [], // TODO: ideally this is encoded as an empty object, consider adding a helper function to encode empty objects
```
dtfiedler, line 100: not really a big concern, we could just set them as nil if we prefer

tests/epochs.test.mjs
```diff
@@ -8,148 +8,337 @@ import {
...
     });
-  });
 
-  describe('Prescribed Names', () => {
-    it('should return the correct epoch for the first epoch', async () => {
-      const prescribedNames = await getPrescribedNames({
-        memory: sharedMemory,
-        timestamp: firstEpochStartTimestamp,
+    describe('Create Next Epoch', () => {
+      it('should create the next epoch at the next epoch start timestamp', async () => {
```
dtfiedler, line 209: new test to validate the creation of the next epoch can occur exactly at its start timestamp

src/demand.lua
```diff
@@ -132,19 +132,17 @@ end
 
 --- Updates the demand factor and returns the updated demand factor
 --- @param timestamp number The current timestamp
---- @return number # The demand factor, updated if necessary
+--- @return number | nil # The demand factor, updated if necessary, nil if no update is necessary
 function demand.updateDemandFactor(timestamp)
 	if not demand.shouldUpdateDemandFactor(timestamp) then
-		print("Not updating demand factor")
-		return demand.getDemandFactor()
+		return nil
```
dtfiedler, line 138: aligns with our other ticking, return `nil` if it's a no-op

src/gar.lua
```diff
@@ -912,6 +912,18 @@ function gar.updateGatewayStats(address, gateway, stats)
 	return gateway
 end
 
+--- Updates the weights for all active gateways. This is usually called every epoch.
+--- @param timestamp number The timestamp to update the weights at
+function gar.updateAllActiveGatewayWeights(timestamp)
```
dtfiedler, line 917: nit: this could be a private/internal function to avoid external callers

tests/helpers.mjs
```diff
@@ -67,7 +70,7 @@ export function assertNoResultError(result) {
     (tag) => tag.name === 'Error',
   );
   assert.strictEqual(errorTag, undefined);
-  assertValidSupplyEventData(result);
+  // assertValidSupplyEventData(result);
```
fedellen, line 73: some trouble in tests here ?

```suggestion
  assertValidSupplyEventData(result);
```

fedellen (approved):
expected a smaller lift here, but looking good

dtfiedler:
> expected a smaller lift here, but looking good

agree - it was originally just going to be stubbed. but my mental model wouldn't let it go. i view these changes similar to how we manage data items in turbo. each step appends new fields and updates various timestamps. it's something that has been rattling in my head for a while.

tests/helpers.mjs
```diff
@@ -895,7 +914,6 @@ export const getRecord = async ({
     },
     memory,
   });
-  assertNoResultError(nameResult);
```
dtfiedler, line 898: these are already handled by `handle` and redundant

src/main.lua
```diff
@@ -1755,16 +1755,6 @@ end, function(msg)
...
 
-	-- if the timestamp is before the genesis epoch start timestamp, do nothing
-	if msg.Timestamp < epochs.getSettings().epochZeroStartTimestamp then
-		Send(msg, {
-			Target = msg.From,
-			Action = "Tick-Notice", -- TODO: this may be better as a "Genesis-Epoch-Not-Started-Notice"
-			Data = json.encode("Genesis epoch has not started yet."),
-		})
-		return
-	end
```
dtfiedler, line 1766: this is handled in `tickEpoch` and redundant

src/main.lua
```diff
@@ -1775,13 +1765,26 @@ end, function(msg)
...
+	-- tick the demand factor
+	local demandFactor = demand.updateDemandFactor(msg.Timestamp)
+	if demandFactor ~= nil then
+		table.insert(newDemandFactors, demandFactor)
+		Send(msg, {
+			Target = msg.From,
+			Action = "Demand-Factor-Updated-Notice",
+			Data = tostring(demandFactor),
+		})
+	end
```
dtfiedler, line 1777: demand factor updates are actually separate from epochs. So they can be pulled up and send their own updates.

src/main.lua
```diff
@@ -2015,27 +2012,6 @@ addEventingHandler(ActionMap.Epoch, utils.hasMatchingTag("Action", ActionMap.Epo
...
-	end
-	table.sort(keys)
-
-	local sortedEpochs = {}
-	for _, key in ipairs(keys) do
-		table.insert(sortedEpochs, allEpochs[key])
-	end
-
-	Send(msg, { Target = msg.From, Action = "Epochs-Notice", Data = json.encode(sortedEpochs) })
-end)
```
dtfiedler, line 2037: removing this handler - it does not scale well and is not needed by anything downstream

tests/tick.test.mjs
```diff
@@ -582,20 +569,21 @@ describe('Tick', async () => {
         totalEpochCount: 1,
       },
       weights: {
-        compositeWeight: 22,
+        compositeWeight: 5.530555555555555,
```
dtfiedler, line 572: i want to confirm these numbers match expectations and will add comments why

dtfiedler:
Note that the tests are marked as failing (1:30PM ET) bc they ran during the distribution delay and no epoch gets returned.

src/epochs.lua
```diff
@@ -87,27 +83,21 @@ EpochSettings = EpochSettings
...
 --- @type Timestamp|nil
 NextEpochsPruneTimestamp = NextEpochsPruneTimestamp or 0
 
---- Gets a deep copy of all the epochs
---- @return table<number, Epoch> # A deep copy of the epochs indexed by their epoch index
-function epochs.getEpochs()
-	return utils.deepCopy(Epochs) or {}
-end
-
+--- TODO: consider removing this handler and prune function below, epochs should manage their own lifecycle
```
arielmelendez, line 91: Still relevant?

src/epochs.lua
```diff
@@ -346,13 +336,12 @@ end
 
 --- Gets the epoch timestamps for an epoch index
 --- @param epochIndex number The epoch index
---- @return number, number, number # 	The epoch start timestamp, epoch end timestamp, and epoch distribution timestamp
+--- @return number, number # 	The epoch start timestamp, epoch end timestamp, and epoch distribution timestamp
```
arielmelendez, line 339: Comment is out of date (still implies a third return value)

arielmelendez: All usages seem to be correctly updated for 2 return values.

src/epochs.lua
dtfiedler: yeah, we needed prune epochs for the initial prune, but should be able to remove it now

src/epochs.lua
```diff
@@ -370,71 +359,62 @@ end
 --- @param timestamp number The timestamp in milliseconds
 --- @param blockHeight number The block height
 --- @param hashchain string The hashchain
---- @return Epoch|nil # The created epoch, or nil if an epoch already exists for the index
-function epochs.createEpoch(timestamp, blockHeight, hashchain)
+--- @return PrescribedEpoch | nil # The created epoch, or nil if an epoch already exists for the index
```
arielmelendez, line 362: Could also qualify that it'd be nil before epoch0 timestamp

src/epochs.lua
```diff
@@ -370,71 +359,62 @@ end
...
+
 	local epochIndex = epochs.getEpochIndexForTimestamp(timestamp)
 	if epochs.getEpoch(epochIndex) then
-		-- silently return
-		print("Epoch already exists for index: " .. epochIndex)
+		print("Epoch already exists for index, skipping creation: " .. epochIndex)
 		return nil -- do not return the existing epoch to prevent sending redundant epoch-created-notices
 	end
 
-	local prevEpochIndex = epochIndex - 1
```
arielmelendez, line 386: Bye, Felicia!

src/gar.lua
```diff
@@ -912,6 +913,8 @@ function gar.updateGatewayStats(address, gateway, stats)
 	return gateway
 end
 
+--- Updates the weights for a gateway
+--- @param weightedGateway WeightedGateway The weighted gateway to update the weights for
 function gar.updateGatewayWeights(weightedGateway)
```
arielmelendez, line 918: Using a copy of the gateway seems pretty clean here, but wondering whether using the unsafe variant of getGateway would also be acceptable...

src/epochs.lua
```diff
@@ -370,71 +359,62 @@ end
...
 	-- get the max rewards for each participant eligible for the epoch
+	local prescribedObservers, updatedGatewaysWithWeights =
+		epochs.computePrescribedObserversForEpoch(epochIndex, hashchain)
+	local prescribedNames = epochs.computePrescribedNamesForEpoch(epochIndex, hashchain)
 	local eligibleEpochRewards = epochs.computeTotalEligibleRewardsForEpoch(epochIndex, prescribedObservers)
-	-- snapshot the ARNS stats at the beginning of the epoch, does not account for any names that are created or expire during the epoch
+	local epochStartTimestamp, epochEndTimestamp = epochs.getEpochTimestampsForIndex(epochIndex)
 	local arnsStatsAtEpochStart = arns.getArNSStatsAtTimestamp(epochStartTimestamp)
+
+	-- update gateway weights for every new epoch, so its clear why the prescribed observers are selected
```
arielmelendez, line 388: Not sure what the assignment here has to do with "clarity" 🤔

arielmelendez: Does hoisting the weights assignment here before the epoch initialization below change any behavior?

src/epochs.lua
```diff
@@ -446,25 +426,30 @@ end
 --- @param observerAddress string The observer address
 --- @param reportTxId string The report transaction ID
 --- @param failedGatewayAddresses table<GatewayAddress> The failed gateway addresses
+--- @param epochIndex number The epoch index
 --- @param timestamp number The timestamp
 --- @return Observations # The updated observations for the epoch
-function epochs.saveObservations(observerAddress, reportTxId, failedGatewayAddresses, timestamp)
+function epochs.saveObservations(observerAddress, reportTxId, failedGatewayAddresses, epochIndex, timestamp)
```
arielmelendez, line 432: This seems like a good update - i.e. ensure that the reported epochIndex has arrived at an appropriate timestamp. Is it possible we weren't enforcing this adequately before? e.g. allowing a previous epoch's report to be submitted in a later epoch?

src/epochs.lua
dtfiedler: > Does hoisting the weights assignment here before the epoch initialization below change any behavior?
it does not
> Not sure what the assignment here has to do with "clarity" 🤔
can reword - but the freshly computed weights are saved in state, and viewable by operators for the epoch

src/epochs.lua
dtfiedler: yeah, that's what caught my eye. I believe `Timestamp` is a protected tag by the MU, so it'll throw it out/error if you try and stub it. but the way we were computing the boundaries for the epoch seemed lacking.

src/epochs.lua
```diff
@@ -586,6 +572,7 @@ function epochs.computeTotalEligibleRewardsForEpoch(epochIndex, prescribedObserv
 		end
 	end
 	return {
+		totalEligibleGateways = #activeGatewayAddresses,
```
arielmelendez, line 575: The function `getActiveGatewaysBeforeTimestamp` could use some luadocs to indicate that an array is returned (and, thus, that # is safe to use here).

src/epochs.lua
```diff
@@ -599,51 +586,55 @@ end
 --- 4. Increment the epoch stats for the gateways
 --- @param currentTimestamp number The current timestamp
 --- @return DistributedEpoch | nil # The updated epoch with the distributed rewards, or nil if no rewards were distributed
-function epochs.distributeRewardsForEpoch(currentTimestamp)
-	local epochIndex = epochs.getEpochIndexForTimestamp(currentTimestamp - epochs.getSettings().durationMs) -- go back to previous epoch
-	local epoch = epochs.getEpoch(epochIndex)
-	if not epoch then
+function epochs.distributeLastEpoch(currentTimestamp)
```
arielmelendez, line 589: In the unusual case that one or more entire epochs have been skipped, "last" might be a misnomer. Perhaps this should be parameterized by epochIndex rather than currentTimestamp?

src/epochs.lua
```diff
@@ -779,9 +775,10 @@ function convertPrescribedEpochToDistributedEpoch(epoch, currentTimestamp, distr
 	}
 end
 
+--- TODO: consider removing this, epochs should manage their own lifecycle and not require prune
```
arielmelendez, line 778: How will the new pattern change the way event data is reported around pruning?

dtfiedler: the distribution/creation should provide the event data on the distributed/new epoch

src/main.lua
```diff
@@ -22,10 +22,11 @@ Vaults = Vaults or {}
 GatewayRegistry = GatewayRegistry or {}
 NameRegistry = NameRegistry or {}
 Epochs = Epochs or {}
+
+-- last known timestamps used to validate behavior to perform on new messages
```
arielmelendez, line 26: Nit: they're not all timestamps

src/main.lua
```diff
@@ -22,10 +22,11 @@ Vaults = Vaults or {}
...
+LastCreatedEpochIndex = LastCreatedEpochIndex or -1 -- TODO: consider making epochs 1 indexed and start at 0
```
arielmelendez, line 27: Not too much more time to decide on the TODO. If we make an adjustment here it likely would be disruptive in testnet but could work fine with a fresh slate in mainnet.

src/main.lua
```diff
@@ -1629,12 +1628,13 @@ end)
 addEventingHandler(ActionMap.SaveObservations, utils.hasMatchingTag("Action", ActionMap.SaveObservations), function(msg)
 	local reportTxId = msg.Tags["Report-Tx-Id"]
 	local failedGateways = utils.splitAndTrimString(msg.Tags["Failed-Gateways"], ",")
+	local epochIndex = msg.Tags["Epoch-Index"]
```
arielmelendez, line 1631: Does the presence of this need to be asserted? Seems like it should because otherwise it will be used for a gte comparison in saveObservations (maybe should assert it there as well).

src/main.lua
```diff
@@ -1775,13 +1765,26 @@ end, function(msg)
 
 	print("Ticking from " .. lastCreatedEpochIndex .. " to " .. targetCurrentEpochIndex)
 
+	-- tick the demand factor
+	local demandFactor = demand.updateDemandFactor(msg.Timestamp)
+	if demandFactor ~= nil then
+		table.insert(newDemandFactors, demandFactor)
```
arielmelendez, line 1771: Will this work correctly in the case where multiple epochs were ticked?

src/main.lua
```diff
@@ -1813,6 +1809,7 @@ end, function(msg)
 				Data = json.encode(tickResult.maybeDistributedEpoch),
 			})
 		end
+		print("Successfully ticked epoch " .. i)
```
arielmelendez, line 1812: Should the reader expect this to mean `i` has been closed out? Or started? Might make more sense to use a verb other than "ticked".

tests/epochs.test.mjs
```diff
@@ -8,149 +8,279 @@ import {
...
+                  operatorReward: 24992562161, // 0.001 of the protocol balance after the transfers and name purchase
+                },
+              },
+            },
+          },
+        });
+        sharedMemory = tickMemory;
       });
     });
+    // TODO: add tests to create N epochs
```
arielmelendez, line 284: For this PR?

src/main.lua
dtfiedler: agree - i think testnet would just "jump an epoch", but will ponder it more and put it for a separate PR

src/main.lua
dtfiedler: demand factor has no affiliation to epochs - it has it's own period length that happens to be the same. `updateDemandFactor` handles that iteration based on the timestamp, so it'll iterate through periods to the current timestamp accordingly.

src/main.lua
dtfiedler: No - bc we may have just created the epoch and not distributed it yet

tests/epochs.test.mjs
dtfiedler: i'll add a test to "skip" epochs for if we ever fall behind

src/epochs.lua
dtfiedler: removed in https://github.com/ar-io/ar-io-network-process/pull/353/commits/11b9e1c33bb0975520a93ec5193b074670982a3f

src/epochs.lua
dtfiedler: updated comment in https://github.com/ar-io/ar-io-network-process/pull/353/commits/11b9e1c33bb0975520a93ec5193b074670982a3f

src/epochs.lua
dtfiedler: added checks for < 1 in https://github.com/ar-io/ar-io-network-process/pull/353/commits/11b9e1c33bb0975520a93ec5193b074670982a3f

src/gar.lua
dtfiedler: updated in https://github.com/ar-io/ar-io-network-process/pull/353/commits/11b9e1c33bb0975520a93ec5193b074670982a3f

src/epochs.lua
dtfiedler: added doc in https://github.com/ar-io/ar-io-network-process/pull/353/commits/11b9e1c33bb0975520a93ec5193b074670982a3f

src/epochs.lua
dtfiedler: removed pruning of epochs in https://github.com/ar-io/ar-io-network-process/pull/353/commits/11b9e1c33bb0975520a93ec5193b074670982a3f

tick should manage epochs in state

src/main.lua
dtfiedler: i made the move to 1 based indexing - updates in https://github.com/ar-io/ar-io-network-process/pull/353/commits/11b9e1c33bb0975520a93ec5193b074670982a3f

I believe for testnet this should just "skip an epoch", we can validate in devnet

src/main.lua
dtfiedler: added checks in https://github.com/ar-io/ar-io-network-process/pull/353/commits/11b9e1c33bb0975520a93ec5193b074670982a3f

src/main.lua
dtfiedler: removed this log as it is unclear. epochs can be in various states - the respective functions add logs about their behavior and outputs - https://github.com/ar-io/ar-io-network-process/pull/353/commits/11b9e1c33bb0975520a93ec5193b074670982a3f

src/main.lua
arielmelendez: Hmm this commit it not found.

arielmelendez: I see it in 11b9e1c3

spec/epochs_spec.lua
```diff
@@ -134,7 +122,7 @@ describe("epochs", function()
 			local expectation = {
 				["observerAddress"] = "test-this-is-valid-arweave-wallet-address-1",
 			}
-			local prescribedObserverMap = epochs.computePrescribedObserversForEpoch(0, hashchain)
+			local prescribedObserverMap = epochs.computePrescribedObserversForEpoch(1, hashchain)
```
arielmelendez, line 125: Why did a bunch of these tests change to using index 1?

src/epochs.lua
```diff
@@ -87,28 +83,15 @@ EpochSettings = EpochSettings
...
-function epochs.getEpochsUnsafe()
-	return Epochs or {}
-end
-
 --- Gets an epoch by index
 --- @param epochIndex number The epoch index
---- @return Epoch | PrescribedEpoch | DistributedEpoch | nil # The prescribed epoch
+--- @return  PrescribedEpoch | nil # The prescribed epoch
 function epochs.getEpoch(epochIndex)
+	if epochIndex < 0 then
```
arielmelendez, line 92: This would be invariant and might be better to assert rather than return nil which is normal behavior in non-invariant scenarios.

arielmelendez: Doing so would allow you to remove the conditionals in all the other getters that fetch the epoch first as they would assert when calling this.

src/epochs.lua
```diff
@@ -344,18 +351,20 @@ function epochs.computePrescribedObserversForEpoch(epochIndex, hashchain)
...
---- Gets the epoch timestamps for an epoch index
+--- Gets the epoch timestamps for an epoch index. Epochs are 0-indexed.
 --- @param epochIndex number The epoch index
---- @return number, number, number # 	The epoch start timestamp, epoch end timestamp, and epoch distribution timestamp
+--- @return number, number # 	The epoch start timestamp, epoch end timestamp
 function epochs.getEpochTimestampsForIndex(epochIndex)
+	if epochIndex < 0 then
+		return 0, 0
+	end
 	local epochStartTimestamp = epochs.getSettings().epochZeroStartTimestamp
```
arielmelendez, line 361: All of these calls to getSettings make a deep copy of the settings. Consider introducing an unsafe variant.

src/epochs.lua
```diff
@@ -476,6 +482,7 @@ function epochs.saveObservations(observerAddress, reportTxId, failedGatewayAddre
 	local observingGateway = gar.getGateway(gatewayAddressForObserver)
 	assert(observingGateway, "The associated gateway not found in the registry.")
 
+	-- TODO: this could be unsafe to avoid copying the epoch on every observation
```
arielmelendez, line 485: Seems like that would be fine.

src/epochs.lua
```diff
@@ -779,39 +814,6 @@ function convertPrescribedEpochToDistributedEpoch(epoch, currentTimestamp, distr
...
-			if endTimestamp >= timestamp then
-				epochs.scheduleNextEpochsPrune(endTimestamp)
-			end
-		end
-		nextEpochIndex = next(unsafeEpochs, nextEpochIndex)
-	end
-	return prunedEpochIndexes
-end
-
 function epochs.scheduleNextEpochsPrune(timestamp)
```
arielmelendez, line 817: Excise

arielmelendez: As well as `nextEpochsPruneTimestamp`

src/epochs.lua
```diff
@@ -597,53 +615,65 @@ end
...
---- @param currentTimestamp number The current timestamp
+--- @param epochIndexToDistribute number The epoch to distribute
 --- @return DistributedEpoch | nil # The updated epoch with the distributed rewards, or nil if no rewards were distributed
-function epochs.distributeRewardsForEpoch(currentTimestamp)
-	local epochIndex = epochs.getEpochIndexForTimestamp(currentTimestamp - epochs.getSettings().durationMs) -- go back to previous epoch
-	local epoch = epochs.getEpoch(epochIndex)
-	if not epoch then
+function epochs.distributeEpoch(epochIndexToDistribute, currentTimestamp)
+	if epochIndexToDistribute < 0 then
+		-- silently ignore - Distribution can only occur after the epoch has ended
```
arielmelendez, line 622: Might make sense to print in case we need to debug.

arielmelendez (approved):

PR #354, opened 2025-02-05
PE-7595: track memory usage at conclusion of handling and gc

arielmelendez:


tests/chaos.test.mjs
dtfiedler, line 1: intended?

dtfiedler (approved):

tests/chaos.test.mjs
arielmelendez: Nope. Removed.

PR #366, opened 2025-02-10
fix(arns): use updated pricing for undername limit purchases, update …

dtfiedler:
…tests

This was missed when we removed the years from undername pricing.

### DIFF
--- a/src/arns.lua
+++ b/src/arns.lua
@@ -290,48 +290,38 @@ end
 --- @param msgId string The current message id
 --- @param fundFrom string|nil The intended payment sources; one of "any", "balance", or "stake". Default "balance"
 --- @return RecordInteractionResult # The result
-function arns.increaseundernameLimit(from, name, qty, currentTimestamp, msgId, fundFrom)
+function arns.increaseUndernameLimit(from, name, qty, currentTimestamp, msgId, fundFrom)
 	fundFrom = fundFrom or "balance"
 	-- validate record can increase undernames
 	local record = arns.getRecord(name)
 
 	assert(record, "Name is not registered")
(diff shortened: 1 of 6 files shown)


tests/arns.test.mjs
```diff
@@ -257,11 +229,23 @@ describe('ArNS', async () => {
...
+
+    // before(async () => {
+    //   const { result: getCostDetailsResult } = await getCostDetails({
+    //     name: 'test-name',
+    //     processId: ''.padEnd(43, 'a'),
+    //     type: 'lease',
+    //     years: 1,
+    //     timestamp: STUB_TIMESTAMP,
+    //   });
+    // });
```
dtfiedler, line 247: adding tests to get the token cost before, and validate balance after purchase

src/arns.lua
```diff
@@ -290,48 +290,38 @@ end
...
-	balances.increaseBalance(ao.id, totalFee)
-	demand.tallyNamePurchase(totalFee)
+	balances.increaseBalance(ao.id, increaseUndernameCost.tokenCost)
+	demand.tallyNamePurchase(increaseUndernameCost.tokenCost)
 	return {
 		record = arns.getRecord(name),
-		totalFee = totalFee,
-		baseRegistrationFee = baseRegistrationFee,
+		totalFee = increaseUndernameCost.tokenCost,
+		baseRegistrationFee = demand.baseFeeForNameLength(#name),
```
dtfiedler, line 324: this is the bug fix - everything else is updating/adding to tests to validate this behaves as expected

arielmelendez (approved):

PR #369, opened 2025-02-11
fix(epochs): remove *eligible rewards* from distributions in epochs handler

dtfiedler:


dtfiedler:
we're going to change this and keep the TOTALs returned by this handler. Eligible recipents will get their own handler

fedellen (approved):

fedellen:
confirmed we can get rid of this, network portal does not use `eligible`:
image

PR #381, opened 2025-02-13
feat: reject data payloads of over 100 bytes

fedellen:


tests/prune.test.mjs
```diff
@@ -46,6 +47,27 @@ describe('ARNS Record Pruning', () => {
     sharedMemory = transferMemory;
   });
 
+  it('messages with more than 100 bytes of data will throw an error', async () => {
```
atticusofsparta, line 50: I think worthwhile to have a test that we can eval large messages

arielmelendez (approved):
Agree with Atticus's comment about the eval test.

tests/prune.test.mjs
```diff
@@ -46,6 +47,27 @@ describe('ARNS Record Pruning', () => {
...
+        Data: longMessage,
+        Tags: [
+          {
+            name: 'Action',
+            value: 'Info',
+          },
+        ],
+      },
+      memory: sharedMemory,
+    });
```
dtfiedler, line 63: nit: move to helpers.mjs

fedellen: [bf39b18](https://github.com/ar-io/ar-io-network-process/pull/381/commits/bf39b1855ac711d24a7312de02110971a3eb01b9)

tests/prune.test.mjs
fedellen: [4d04db8](https://github.com/ar-io/ar-io-network-process/pull/381/commits/4d04db8bf5fa3d0d46a59411f65d70baff1fc895)

tests/prune.test.mjs
```diff
@@ -46,6 +47,34 @@ describe('ARNS Record Pruning', () => {
     sharedMemory = transferMemory;
   });
 
+  it('Eval action messages with more than 100 bytes of data will not throw an error', async () => {
+    const { result } = await sendEval({
+      memory: sharedMemory,
+      data: '-- a'.repeat(101) + "\nprint('hello')",
```
arielmelendez, line 53: Doesn't test the exact boundary condition but is good enough for now.

arielmelendez (dismissed):

src/main.lua
```diff
@@ -420,6 +420,11 @@ end
 
 -- Sanitize inputs before every interaction
 local function assertAndSanitizeInputs(msg)
+	if msg.Tags.Action ~= "Eval" and msg.Data then
+		-- assert data size is less than or equal to 100 bytes, we dont use msg.Data
+		assert(#msg.Data <= 100, "Data size is too large")
```
atticusofsparta, line 425: if the content type is JSON and a json string is provided, this could fail. Recommend adding a test specifically that large tx's with a json content type fail

fedellen: updated! nice catch

src/main.lua
```diff
@@ -420,6 +420,16 @@ end
 
 -- Sanitize inputs before every interaction
 local function assertAndSanitizeInputs(msg)
+	if msg.Tags.Action ~= "Eval" and msg.Data then
+		-- assert data size is less than or equal to 100 bytes, we dont use msg.Data
+
+		if type(msg.Data) == "table" then
+			msg.Data = json.encode(msg.Data)
```
dtfiedler, line 427: the json.encode here with a large data size will kill us - we want to essentially avoid doing this entirely.

src/main.lua
dtfiedler: @atticusofsparta why would this fail if msg.Data is a stringiifed json? we want to avoid encoding if we don't need to

fedellen: 42b1432

src/main.lua
fedellen: 42b1432

src/main.lua
fedellen: #msg.Data on lua type of table was indeed not asserting correctly when Content-Type application/json is fed through AOS

atticusofsparta: @dtfiedler in the top level global handle, if the content type is set to application/json, it will decode the msg.Data field and overwrite it with a lua table.

src/main.lua
atticusofsparta: > the json.encode here with a large data size will kill us - we want to essentially avoid doing this entirely.

nit: If it is a large data size and we throw then it shouldn't matter, no? JSON encoding sub 100 bytes is trivial, might be worth it for simplicity?

src/main.lua
atticusofsparta: Now, it wouldn't actually break, but it would be inaccurate. For example if it was a table with 1 element, it would have a length of 1, but not a byte length of 1

src/main.lua
```diff
@@ -418,8 +418,47 @@ local function addPrimaryNameRequestData(ioEvent, primaryNameResult)
...
+		if current_size > max_bytes then
+			error("Data size exceeds " .. max_bytes .. " bytes")
+		end
+
+		local t = type(val)
+		if t == "string" then
+			current_size = current_size + #val
+		elseif t == "number" or t == "boolean" then
+			current_size = current_size + 8 -- Approximate size for numbers/booleans
+		elseif t == "table" and not seen[val] then
```
arielmelendez, line 435: I'd be pretty surprised if the json deserializer is smart enough to save on table references in this way but nice thoroughness here.

src/main.lua
```diff
@@ -418,8 +418,47 @@ local function addPrimaryNameRequestData(ioEvent, primaryNameResult)
 	addPrimaryNameCounts(ioEvent)
 end
 
+local function assertTableBytesLowerThan(tbl, max_bytes)
+	local seen = {}
+	local current_size = 0
+
+	local function sizeof(val)
+		if current_size > max_bytes then
```
arielmelendez, line 426: Shouldn't this be at the end of the function? If it were I think you could remove it from the table section.

src/main.lua
```diff
@@ -418,8 +418,47 @@ local function addPrimaryNameRequestData(ioEvent, primaryNameResult)
...
+
+	sizeof(tbl)
+end
+
 -- Sanitize inputs before every interaction
 local function assertAndSanitizeInputs(msg)
+	if msg.Tags.Action ~= "Eval" and msg.Data then
+		-- assert data size is less than or equal to 100 bytes, we dont use msg.Data
+
+		if type(msg.Data) == "table" then
```
arielmelendez, line 455: This conditional would go away with the generalization of `assertValueBytesLowerThan`.

arielmelendez (approved):

src/main.lua
fedellen: [f77a84f](https://github.com/ar-io/ar-io-network-process/pull/381/commits/f77a84f75d612296705b4d41f4dcc3ab175667e9)

src/main.lua
```diff
@@ -418,8 +418,47 @@ local function addPrimaryNameRequestData(ioEvent, primaryNameResult)
 	addPrimaryNameCounts(ioEvent)
 end
 
+local function assertTableBytesLowerThan(tbl, max_bytes)
```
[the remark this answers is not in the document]
fedellen: [f77a84f](https://github.com/ar-io/ar-io-network-process/pull/381/commits/f77a84f75d612296705b4d41f4dcc3ab175667e9)

src/main.lua
fedellen: fwiw I believe json.encode happens within the AO code -- but our size checking function works here fine

src/main.lua
```diff
@@ -418,45 +418,32 @@ local function addPrimaryNameRequestData(ioEvent, primaryNameResult)
...
-			seen[val] = true
-			for k, v in pairs(val) do
-				sizeof(k)
-				sizeof(v)
-				if current_size > max_bytes then
-					error("Data size is too large")
-				end
-			end
+local function assertValueBytesLowerThan(value, remainingBytes, tables_seen)
+	local tablesSeen = tables_seen or {}
```
arielmelendez, line 422: nit: it's safe to just reassign into tables_seen arg:
```
tables_seen = tables_seen or {}
```

arielmelendez (dismissed):

PR #385, opened 2025-02-14
feat: set demand factor to 2 and adjust test expectations

fedellen:


tests/tick.test.mjs
```diff
@@ -727,15 +727,17 @@ describe('Tick', async () => {
...
 
     // compute the periods until we get to 0.5 from the current demand factor
-    const periodsUntilMinDemandFactor = Math.ceil(
-      Math.log(demandFactorSettings.demandFactorMin) /
-        Math.log(1 - demandFactorSettings.demandFactorDownAdjustmentRate),
-    );
+    const periodsUntilMinDemandFactor =
+      Math.ceil(
+        Math.log(demandFactorSettings.demandFactorMin) /
+          Math.log(1 - demandFactorSettings.demandFactorDownAdjustmentRate),
```
dtfiedler, line 738: nit: this 1 can be the DemandFactor.currentDemandFactor. That would remove the need for the * 2 and should work with any initial demand factor value

tests/tick.test.mjs
```diff
@@ -727,15 +727,17 @@ describe('Tick', async () => {
...
+      ) * 2; // we multiply by 2 because we start with 2 demand factor
+    console.log('periodsUntilMinDemandFactor', periodsUntilMinDemandFactor);
```
dtfiedler, line 740:
```suggestion
```

dtfiedler (commented):
not: for all of these, instead of using * 2, we can use getDemandFactor helper to fetch the demand factor from the contract

PR #392, opened 2025-02-17
fix(PE-7709): update primary names to support undernames

atticusofsparta:


src/arns.lua
```diff
@@ -700,12 +703,14 @@ end
...
 --- @return TokenCostResult tokenCostResult The token cost result of the intended action
 function arns.getTokenCost(intendedAction)
+	local intent = intendedAction.intent
 	local tokenCost = 0
 	local purchaseType = intendedAction.purchaseType or "lease"
 	local years = tonumber(intendedAction.years)
 	local name = intendedAction.name
-	local baseFee = demand.baseFeeForNameLength(#name)
-	local intent = intendedAction.intent
+	local baseFee = demand.baseFeeForNameLength(
```
dtfiedler, line 711: you can see on line 761 we have this value set, not necessary to add this conditional here

atticusofsparta: We need to add it because theres an assert in baseFeeForNameLength that it retrieved a fee - if we have a 51 char base name + an undername, this will attempt to access an out of bounds fee (eg length of 52 +) and throw an error.

One option was to use the getBaseName util, this seemed better at the time. Thoughts?

dtfiedler: yeah, lets use the getBaseName util and add a comment why it's necessary

atticusofsparta: [e667966](https://github.com/ar-io/ar-io-network-process/pull/392/commits/e66796656ad100448e4db4f7aabe0678b6db5e9b)

src/primary_names.lua
```diff
@@ -50,8 +50,60 @@ NextPrimaryNamesPruneTimestamp = NextPrimaryNamesPruneTimestamp or 0
...
-local function baseNameForName(name)
-	return (name or ""):match("[^_]+$") or name
+-- NOTE: lua 5.3 has limited regex support, particularly for lookaheads and negative lookaheads or use of {n}
+---@param name string
+---@description Asserts that the provided name is a valid undername
+---@example
+---```lua
+---utils.assertValidateUndername("my-undername")
+---```
+function primaryNames.assertValidUndername(name)
```
dtfiedler, line 60: nit - assertValidPrimaryName

src/primary_names.lua
```diff
@@ -50,8 +50,60 @@ NextPrimaryNamesPruneTimestamp = NextPrimaryNamesPruneTimestamp or 0
...
+	assert(name and type(name) == "string", "Name is required and must be a string.")
+
+	assert(
+		#name <= constants.MAX_PRIMARY_NAME_LENGTH,
+		"Primary Name is too long, received length of " .. tostring(#name)
+	)
+
+	-- Ensure name starts and ends with alphanumeric, and only allows `_` or `-` in between
+	assert(
+		string.match(name, "^[a-zA-Z0-9]$") ~= nil
```
dtfiedler, line 88: put this regex into constants file

dtfiedler: and please confirm with GPT that lua supports these unioned regexes - there may be a more concise regex available

spec/primary_names_spec.lua
```diff
@@ -156,6 +211,60 @@ describe("Primary Names", function()
...
+			"1_test",
+			"1234_test",
+			"fsakdjhflkasjdhflkaf_test",
+			("").rep("1", 61) .. "_t",
+			"a_" .. ("").rep("1", 51),
+		}
+
+		for _, primaryName in ipairs(validPrimaryUndernameNames) do
+			it(
+				"should create a primary name request with undername and transfer the cost from the initiator to the protocol balance",
```
arielmelendez, line 225: It's not clear to me that unit testing this function over an array of valid undernames is testing for the right assertions, or at least not doing so in a a straightforward way. I think this makes it harder to evaluate your added code for correctness.

arielmelendez: Said differently, I'd have preferred to see unit tests specifically for, e.g.:
- utils.baseNameForName
- arns.getTokenCost
- assertValidUndername

and also vectorizing the happy path tests for affected primary names functions with their undername counterparts.

arielmelendez: Alternatively, I'd take a few integration test updates.

arielmelendez: Your integration test coverage probably exercises enough of what we need. Would hurt to pick off some of the low hanging fruit from the above list though.

atticusofsparta: [13595b5](https://github.com/ar-io/ar-io-network-process/pull/392/commits/13595b58a4d30cddcb54fc48a39090301b2d42fa)

dtfiedler (dismissed):

spec/utils_spec.lua
```diff
@@ -934,4 +934,14 @@ describe("utils", function()
 			assert.is_false(utils.booleanOrBooleanStringToBoolean(1))
 		end)
 	end)
+	describe("baseNameForName", function()
+		it("should get base name for name with an undername", function()
```
arielmelendez, line 938: Let's test this a bit more aggressively now that we're leaning on it even more.

fedellen (approved):

dtfiedler (approved):

arielmelendez (approved):

fedellen (commented):
✅

PR #396, opened 2025-02-18
chore(crypto): remove local copy of crypto

dtfiedler:
Crypto is provided by the module, and can be referenced via .crypto.init

Moves `crypto` to spec to keep unit tests working, but otherwise strips it out from the source code


spec/setup.lua
```diff
@@ -21,6 +21,10 @@ _G.Handlers = {
 	},
 }
 
+-- Then stash it in package.loaded under the name ".crypto.init"
```
arielmelendez, line 24: Nit: disjointed comment? (why "Then")?

arielmelendez (approved):

PR #398, opened 2025-02-18
chore: devnet to testnet

fedellen:


dtfiedler:
FYI - this will override the block in - https://github.com/ar-io/ar-io-network-process/pull/394

dtfiedler (approved):

arielmelendez (approved):

PR #403, opened 2025-02-20
fix(paginated vaults): use vaultId for a unique cursor

fedellen:


tests/vaults.test.mjs
```diff
@@ -611,7 +612,15 @@ describe('Vaults', async () => {
         from: secondVaulter,
         msgId: vaultId2,
       });
-      paginatedVaultMemory = updatedMemory2;
+      const { memory: updatedMemory3 } = await createVault({
```
dtfiedler, line 615: nit - have this a createVaultedTransfer and assert the controller is returned as well

arielmelendez (approved):

dtfiedler (approved):

PR #405, opened 2025-03-03
chore(patches): add mainnet patches

dtfiedler:


patches/2025-02-20-pagination-cursor-field.lua
```diff
@@ -0,0 +1,96 @@
...
+			nextCursor = nil,
+			hasMore = false,
+		}
+	end
+
+	local startIndex = 1
+
+	if cursor then
+		for i, obj in ipairs(sortedArray) do
+			if cursorField and obj[cursorField] == cursor or cursor == obj then
```
arielmelendez, line 45: Hmm reviewing this again I now worry about what happens if obj is not indexable as a table and we attempt to index it by the cursorField 🤔

PR #406, opened 2025-03-04
fix(tick): fix event data on tick

dtfiedler:
Properly record the observer addresses and new demand factor periods and values on tick

### DIFF
--- a/patches/2025-03-04-fix-observer-event-data.lua
+++ b/patches/2025-03-04-fix-observer-event-data.lua
@@ -1,0 +1,180 @@
+--[[
+	Fixes a bug where the prescribed observers were not being added to event data on tick.
+
+	Related PR: https://github.com/ar-io/ar-io-network-process/pull/406
+
+	Reviewers: Dylan, Ariel, Derek, Jonathon, Phil
+]]
+--
+
+local utils = require(".src.utils")
+
+-- Find reference to the handler for the distribute action in the Handlers table
+local distributeHandlerIndex = utils.findInArray(Handlers.list, function(handler)
+	return handler.name == "distribute"
+end)
+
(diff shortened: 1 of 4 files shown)


fedellen (dismissed):

src/ao_event.lua
dtfiedler, line 1: this is a bit harder to patch, since it's a global class referencing an object only on instantiation. we'd need to essentially replace the entire object, which we can do - just makes the patch large

patch.mjs
```diff
@@ -0,0 +1,26 @@
...
+const patchName = process.argv[2];
+
+if (!patchName) {
+  console.error('Patch name is required');
+  process.exit(1);
+}
+
+fs.mkdirSync('patches', { recursive: true });
+fs.writeFileSync(
+  `patches/${date}-${patchName.replace(/ /g, '-').toLowerCase().trim()}.lua`,
```
arielmelendez, line 24: We'll have to introduce any stateful versioning info on the input patch name side here in the event that we have multiple versions of a patch on the same day (not a big deal at all. just making sure we're aligned on that).

patches/2025-03-04-fix-observer-event-data.lua
```diff
@@ -0,0 +1,180 @@
...
+	Related PR: https://github.com/ar-io/ar-io-network-process/pull/406
+
+	Reviewers: Dylan, Ariel, Derek, Jonathon, Phil
+]]
+--
+
+local utils = require(".src.utils")
+
+-- Find reference to the handler for the distribute action in the Handlers table
+local distributeHandlerIndex = utils.findInArray(Handlers.list, function(handler)
```
arielmelendez, line 13: Do we definitely have access to Handlers?

PR #414, opened 2025-03-27
fix(PE-7752): fix observer event data

dtfiedler:
The event data using tables does not correctly get added to honeycomb, making it hard for us to track observers selected over time.


src/ao_event.lua
```diff
@@ -60,7 +60,9 @@ local function AOEvent(initialData)
 		end
 		if not isValidType(value) then
 			print(
-				"ERROR: Invalid field value type: "
+				"ERROR: Invalid field value type for key ("
+					.. key
+					.. "): "
```
dtfiedler, line 65: we'll need to patch this in and replace the existing global scope with this updated `AOEvent` - this patch is not currently included in this PR - but will add once other pieces are tested

fedellen (approved):

PR #415, opened 2025-03-31
chore(PE-7894): add cranking monitor for transfer messages

atticusofsparta:
image


tests/monitor/cranking-monitor.test.mjs
```diff
@@ -0,0 +1,197 @@
+import { connect } from '@permaweb/aoconnect';
+//import { DockerComposeEnvironment, Wait } from 'testcontainers';
+
+// const projectRootPath = process.cwd();
+
```
dtfiedler, line 5: nit: remove these comments

dtfiedler (commented):
some cleanup to do before merge

tests/monitor/cranking-monitor.test.mjs
atticusofsparta: https://github.com/ar-io/ar-io-network-process/pull/415/commits/0c7f163a272596181eb82244fb2e9719846ec2ad

.github/workflows/cranking.yaml
```diff
@@ -0,0 +1,133 @@
...
+              exit 0
+            else
+              echo "Attempt $i failed with exit code $exit_code"
+              if [ $i -lt 4 ]; then
+                echo "Retrying in 15 seconds..."
+                sleep 15
+              fi
+            fi
+          done
+          echo "All 4 attempts failed"
```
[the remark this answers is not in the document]
atticusofsparta: I removed the retries - I dont think we really capture the failed id's on the notifications, would like to add those ideally, looking at what to do for that now.

dtfiedler (approved):

PR #418, opened 2025-04-02
fix(PE-7908): update resolution pattern for cranking status

atticusofsparta:


dtfiedler (approved):

tests/monitor/cranking-monitor.test.mjs
```diff
@@ -35,90 +41,224 @@ async function getMessagesForProcessByAction({
     .map((edge) => edge.node.message.id);
 }
 
+// returns a list of ids that are processes
+async function getProcessMetadataFromGql({ processIds }) {
+  let attempts = 0;
+  const maxAttempts = 10;
+  const backoffTime = 1000; // 1 second
```
dtfiedler, line 48: nit: use exponential back off

atticusofsparta: yup, its handled in the catch

PR #419, opened 2025-04-03
fix(cranking monitor): add content type header to gql request

atticusofsparta:


dtfiedler (approved):
what will this do?

atticusofsparta:
@dtfiedler well apparently the post request wasnt configured properly, and it wasnt reading the gql body because of it. I found out by attempting to post to goldsky and got an error regarding no body (despite there being one).

PR #434, opened 2025-09-17
fix(undername util): fix undernameForName to handle dashes in the base name

atticusofsparta:


spec/utils_spec.lua
```diff
@@ -1027,13 +1027,55 @@ describe("utils", function()
...
+			assert.are.same(undername, undernameFromName)
+		end)
+
+		it("should return nil for a base name with dashes but no undername", function()
+			local basename = "base-name-with-dashes"
+			local undernameFromName = utils.undernameForName(basename)
+			assert.are.same(nil, undernameFromName)
+		end)
+
+		it("should handle complex undername with base name containing dashes", function()
```
arielmelendez, line 1054: Nit: "should handle" is not descriptive enough of the intended behavior so it's harder for a reader to understand whether the test is implemented appropriately. The tests above do a better job describing the expected result.

src/utils.lua
```diff
@@ -783,7 +783,9 @@ function utils.undernameForName(name)
 	end
 
 	local baseName = utils.baseNameForName(name)
-	return string.gsub(name:reverse(), baseName:reverse() .. "_", "", 1):reverse()
+	-- Escape the pattern to treat it as literal text, not a pattern
+	local pattern = string.gsub(baseName:reverse() .. "_", "([%-%.%+%[%]%(%)%$%^%%%?%*])", "%%%1")
```
arielmelendez, line 787: Readability here is pretty bad and this seems unnecessarily complicated. I've got an idea I'll push for ya now that you've got the tests in place...

arielmelendez: 492fe1b

arielmelendez: And also d120437

arielmelendez (approved):

